PR #57, opened 2023-10-17
Merging security fix & copyedits.

jakimfett:
Doing a PR more to work out how we do this process than anything else.


jakimfett:
It probably makes sense to do branch PRs instead of directly against `main`...thoughts?

michaelthatsit:
@jakimfett If you look at previous PRs I've been branching off of main and PRing to main with every feature or bugfix, and adding `feature/` or `fix/` as a prefix to give a glanceable indicator of what it is.

PR #94, opened 2023-11-09
build/test automation setup

hanbollar:


michaelthatsit (approved):
This looks good to me. Looks like there's some line the got removed in the build file, but it was commented out code and doesn't exist on my branch.

PR #121, opened 2023-11-16
separating dist, assets, && samples

hanbollar:
makes the dist folder 100% cleanable and allows for extension so we can add more samples later on as well

modify the files in assets and samples and they'll be copied to dist on build

~~also ignores keeping dist in main branch anymore since that's just a build folder~~


hanbollar:
going to readd the dist folder and leave as todo for server to render where we serve the dist as part of the process instead of directly from the repo

PR #125, opened 2023-11-16
fix jest/node filepath conflicts so can run tests

hanbollar:
includes testing for glb and stl to start

at some point in future we'll have to get testing setup to also test the loaded result, but for now, testing the process of loading is enough alongside visual check

automated testing in action wont pass due to lingering async note so the action will show red, but in the action and locally the tests themselves will pass


src/core/entity.js
```diff
@@ -315,7 +315,7 @@ export default class Entity extends MRElement {
     const children = Array.from(this.children);
     for (const child of children) {
       // if o is an object, traverse it again
-      if (!child instanceof Entity) {
+      if ((!child) instanceof Entity) {
```
takahirox, line 318: Is this change intentional? `!child` would be boolean (true/false) then ((!child) instanceof Entity) would be always false.

hanbollar: that was leftover from a previous change, good catch - https://github.com/Volumetrics-io/mrjs/pull/521

PR #128, opened 2023-11-17
fix prettier and eslint infighting

hanbollar:
- removed airbnb configs

we should keep `prettier` for just pretty/style
current things we're enforcing:
- es5 trailing comma
- semicolon at line ends
- single quotes instead of double
- tab with of 4 spaces
- no tabs allowed
- print width max: 180

we should use `eslint` to enforce docs/var/grammar-based aspects
current things we're enforcing:
- docs
- no use of 'var' only 'const' or 'let'


hanbollar:
gonna leave this as an open pr if others have pending code since it's a big one

hanbollar:
just adding review for another set of eyes if there's more settings that might be nice - not urgent for merge

PR #148, opened 2023-11-21
WIP/quicksave - trying out accomodations for hair in island.glb

hanbollar:
- glb should just work with the hair - looking into why it doesnt in our current loader
- temp save for now


hanbollar:
not doing for now - leaving for other model loading

Issue #89, opened 2023-11-07
Integrate Skyboxes as a background to a web page


Dinoel27:
Hello:) 

Is it possible to integrate a skybox into MR.js? 

The skybox will appear in the background once the website is opened. The skybox would load/appear in the same direction as the webpage is presented to the user (top to down). The skybox can be toggled on/off. Also, the skybox could be shown as a half dome vertically or horizontally. Users can add multiple skyboxes and scroll through them as they progress on the web page. 

If skybox is a feature of a tool for creating an XR website, users should be able to import the skybox correctly. 

For now, these are all the thoughts I have

michaelthatsit:
@lobau assigning to you since you're already working on it. it'll be cool to integrate this directly into the library 🥳

Issue #84, opened 2023-11-07
full support for css background


michaelthatsit:

hanbollar:
closing in favor of https://github.com/Volumetrics-io/MR.js/issues/167

PR #153, opened 2023-11-22
add skybox texture entity

hanbollar:
related to #89
related to #84


src/entities/SkyBox.js
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     *
+     */
+    connected() {
+        // you can have texturesList be all individual textures
+        // or you can store them in a specified path and just
+        // load them up solely by filename in that path.
+        
+        this.texturesList = this.getAttribute('textures');
```
michaelthatsit, line 14: @hanbollar what do we thing about using `src` instead of textures? We could also use the 'background' css property which would be sick.

src/component-systems/FadeSystem-todo.js
michaelthatsit, line 1: Should this be a system exclusively for fading or part of a larger animation system?

I'd also like to see it generalized to any entity and utilize a `data-comp-fade` (or `data-comp-animation`) rather than just getting all the skyboxes in the scene.

michaelthatsit (commented):
I left some comments, overall pretty good, but not sure I want the commented out fadesystem to make it into main if it's not fully implemented.

src/component-systems/FadeSystem-todo.js
hanbollar: yea this was just a pure copy - hence the commenting it out - i'll just remove it and push into another WIP pr to actually impl something there instead

src/entities/SkyBox.js
hanbollar: sounds good for src 👍  - i'll look into the background css property as well

hanbollar:
@michaelthatsit - lmk if these changes fix what you were requesting - https://github.com/Volumetrics-io/MR.js/pull/153/files/02769a95e66bb67ea8a86d4caf1e0769beb8eda7..55db980f8862aaa7bc8d0333ea289c861b49b2a0

michaelthatsit (approved):

src/entities/SkyBox.js
```diff
@@ -11,20 +11,21 @@ class SkyBox extends Entity {
...
         const textureNames = texturesList.split(',');
-        let texture = null;
 
         let path = this.getAttribute('pathToTextures');
         const texture = !path ? textureLoader.load(textureNames) : textureLoader.load(textureNames.map((name) => path + name));
 
         // scene.background
         this.object3D.background = texture;
+        // css background
+        this.compStyle.background = texture;
```
michaelthatsit, line 28: ah so you wouldn't set compStyle.background really ever. that's done with css. Rather I'd expect you'd use `this.compStyle.background` as an alternative to `src`.

something like:

```
this.texturesList = this.compStyle.backgroundImages
...
const textureNames = texturesList.split([some regex to parse list of paths]);
```

the difference between `src` and `background-image` would be the list of paths. each path in background-image is wrapped with `url("...")` so you'd need to do some regex.

`background-image: url("img_tree.gif"), url("paper.gif");`

michaelthatsit: Long term it would be sick to use every feature of the `background` CSS property (color, image, repeat, position etc)  to define textures.

`background: lightblue url("img_tree.gif") no-repeat fixed center;`

PR #165, opened 2023-11-28
margin fix

hanbollar:
small change - @michaelthatsit can you sanity check for me that this was all that was needed - afaik this was a legacy thing that shouldnt be needed anymore, but lmk if not

related to https://github.com/Volumetrics-io/MR.js/issues/157


michaelthatsit:
What was the issue related to this? This was needed to set margins for ui elements. Last I checked it worked.

michaelthatsit:
Ah ignore the above. I'm on the mobile app while Neha is driving and still getting used to it.

The margin lines aren't legacy and are used in the current index page. Lemme check the existing readme. Pretty sure it's that one of the elements used for that example no longer exist, so compStyle and all of its properties won't exist.

PR #177, opened 2023-12-09
160 layout bugs

michaelthatsit:
lot's of changes, updating readme as well. might rename mr-container.


src/UI/UIEntity.js
```diff
@@ -59,6 +59,35 @@ export class MRUIEntity extends Entity {
...
+    add(entity) {
+        let container = this.closest('mr-panel');
+
+        if (container && entity instanceof MRUIEntity) {
+            container.add(entity);
+        } else {
+            this.object3D.add(entity.object3D);
+        }
+
+        entity.object3D.position.z += 0.0001;
```
hanbollar, line 75: needs comment above this line for why this bump is necessary

src/UI/media/Image.js
```diff
@@ -36,6 +36,8 @@ export class MRImage extends MRUIEntity {
             }
         });
         this.object3D.material.map = this.texture;
+
+        this.object3D.position.z += 0.00001;
```
hanbollar, line 40: needs comment above this line for why this bump is necessary

src/component-systems/TextSystem.js
```diff
@@ -63,15 +63,18 @@ export class TextSystem extends System {
                     entity.blur();
                 }
             } else {
-                text = entity.textContent.replace(/(\n)\s+/g, '$1').trim();
+                text = entity.textContent
+                    .replace(/(\n)\s+/g, '$1')
+                    .replace(/(\r\n|\n|\r)/gm, '')
+                    .trim();
```
hanbollar, line 69: need comment above this explaining why this swap is needed and safe

src/core/entity.js
```diff
@@ -49,8 +50,9 @@ export default class Entity extends MRElement {
      *
      */
     get height() {
-        const styleHeight = this.compStyle.height.split('px')[0] > 0 ? this.compStyle.height.split('px')[0] : window.innerHeight;
-        return (styleHeight / window.innerHeight) * global.viewPortHeight;
+        const rect = this.getBoundingClientRect();
+        // const styleHeight = this.compStyle.height.split('px')[0] > 0 ? this.compStyle.height.split('px')[0] : window.innerHeight;
```
hanbollar, line 54: can this line be deleted?

src/entities/Model.js
```diff
@@ -15,9 +15,9 @@ export class Model extends Entity {
     /**
      *
      */
-    get height() {
-        return this.contentHeight;
-    }
+    // get height() {
+    //     return this.contentHeight;
+    // }
```
hanbollar, line 20: can the above be deleted?

src/entities/Surface.js
```diff
@@ -35,10 +48,12 @@ export class Surface extends Entity {
      *
      */
     connected() {
-        this.windowVerticalScale = this.height / 3;
-        this.windowHorizontalScale = this.width / 3;
+        // this.clipping = new ClippingGeometry(new THREE.BoxGeometry(this.width, this.height, 0.3));
```
hanbollar, line 51: can this line be deleted?

src/entities/Surface.js
```diff
@@ -35,10 +48,12 @@ export class Surface extends Entity {
...
+        this.windowVerticalScale = this.height; /// 3;
+        this.windowHorizontalScale = this.width; // / 3;
```
hanbollar, line 53: question for understanding: (i see this was moved to line 67) is the divide by 3 just a choice or is there a reason we're using 3 specifically?

src/entities/Surface.js
```diff
@@ -111,18 +126,24 @@ export class Surface extends Entity {
         this.object3D.position.copy(this.anchorPosition);
         this.object3D.quaternion.copy(this.anchorQuaternion);
 
+        console.log('floating', this.floating);
+
+        if (!this.floating) {
+            this.rotationPlane.rotation.x = (3 * Math.PI) / 2;
```
hanbollar, line 132: need comment as to why this additional rotation is needed

hanbollar (approved):
some quick notes for comments / straggler lines and a question or two 🎉

README.md
```diff
@@ -43,98 +43,54 @@ Both options require you generate an ssl certificate & key via openssl:
...
-                    <mr-img src="..."></mr-img>
-                    <mr-row height="0.02">
-                        <mr-button onClick="Prev()"> <- </mr-button>
-                        <mr-button onClick="Next()"> -> </mr-button>
-                    </mr-row>
-                </mr-column>
-            </mr-row>
-        </mr-container>
-    </mr-surface>
+    <!-- The 2D UI Panel -->
```
hanbollar, line 80: could we also put one element that is 3d outside of the panel for the readme demo? ik we have included the mr-model elemnt on line 88, just a bit concerned of using just the naming of 2D though

hanbollar (approved):
one quick additional question

src/entities/Surface.js
michaelthatsit: > question for understanding: (i see this was moved to line 67) is the divide by 3 just a choice or is there a reason we're using 3 specifically?

This is a choice atm, the window is really big if we stick to 1:1 scale. I'd like to refactor the anchoring/surface system overall. so we can parameterize this in the future?

README.md
michaelthatsit: ooo good catch. I might add this in another PR with the button/img update. I'd like it to be in the demo page as well.

src/entities/Surface.js
hanbollar: sounds good - then in line 67, either add a comment or can we separate the divide by 3 into a constant first and then divide - better visual separation for future parameterization - just dont want it to end up as a 'secret special number'

michaelthatsit: Made it a global.

michaelthatsit:
@hanbollar good to merge?

PR #187, opened 2023-12-14
185 mr img responsive fix

michaelthatsit:


src/utils/Css.js
```diff
@@ -15,8 +15,14 @@ export function threeToPx(val) {
  * @returns {number} - the 3D representation of value.
  */
 export function pxToThree(val) {
+    let px;
+    if (val instanceof String) {
+        px = val.split('px')[0];
+    } else {
+        px = val;
+    }
```
hanbollar, line 23: the above should be a ternary for better readability

`let px = (val instanceof String) ? val.split('px'[0] : val;`

hanbollar (approved):

src/utils/Css.js
michaelthatsit: good idea.

PR #192, opened 2023-12-15
add masking of mrDivEntity elements by panels

hanbollar:


hanbollar:
tomorrow going to test that it works as the default and then what happens when allowing items to be turned off from it

hanbollar:
putting a pin in this to revisit tomorrowish - potentially requires a much larger architecture change than just some quick mesh / shader passes - storing small version that works in `potential_fix` folder on this pr for now while i rethink the initial mesh grab / separation approach

additional context:
- div entities parenting to the panel regardless of their parent.

i need to look into whether i can take advantage of this or not - tbh given that, it just probably means a refactor of ignoring all the panel children and grabbing them separately before the mask - might not need larger architecture change

hanbollar:
wooooorks, just need to remove debug colorings and allow for multiple panel stencils (small change on top) - incoming ~

michaelthatsit:
Just pulled, I'm noticing a drop in frame rate both in headset and in chrome.  reviewing code now.

hanbollar:
> Just pulled, I'm noticing a drop in frame rate both in headset and in chrome. reviewing code now.

that's partially expected since we're taking more steps in the render loop - i'm doing the for-loop based on the assumption that we'll allow for multiple panels and the panels might even overlap in future, but i can default back to just one main panel/masking ref (ie all panels count as one mask) for future

additionally - i havent looked into future optimizations behind how we're doing all the system updates yet, and there's definitely speed improvements we can do there still

lmk if you'd want me to do the render streamline i mentioned though for this pr

michaelthatsit:
> that's partially expected since we're taking more steps in the render loop - i'm doing the [for-loop](https://github.com/Volumetrics-io/mrjs/pull/192/files#diff-9a9b14ca4cde264f45d74bc2e9e2926a929c81d6fccf4161133d5fc99d48d3faR447) based on the assumption that we'll allow for multiple panels and the panels might even overlap in future, but i can default back to just one main panel/masking ref (ie all panels count as one stencil for all sub objects) for future

hmm but it doesn't look like you're applying that every update? 

it's also not clear if it has anything to do with this PR. I went through and disabled each system to see where the perf hits might be coming from. disabling masking didn't change it.

src/core/MRApp.js
```diff
@@ -410,7 +424,37 @@ export class MRApp extends MRElement {
...
+
+        // TODO (in future) - once this gets more complicated, it will be nice to have a render system separate
+        // from the pure loop but it is okay as is here for now.
+
+        // Need to wait until we have all needed rendering-associated systems loaded.
+        if (this.maskingSystem == undefined) {
+            return;
+        }
+
+        const camera = this.user;
```
michaelthatsit, line 437: Let's just pass `this.user` into the render calls.

src/core/MRApp.js
```diff
@@ -410,7 +424,37 @@ export class MRApp extends MRElement {
...
+
+        // Render panel to stencil buffer
+        this.renderer.state.buffers.stencil.setTest(true);
+        this.renderer.state.buffers.stencil.setMask(0xff);
+        this.renderer.render(this.scene, camera);
+
+        // Render child-objects to where the stencil buffer is set; must start at 1, see MaskingSystem for more details.
+        for (let panel_ref = 1; panel_ref <= this.maskingSystem.panels.length; ++panel_ref) {
+            this.renderer.state.buffers.stencil.setFunc(THREE.EqualStencilFunc, panel_ref, 0xff);
+            this.renderer.render(this.scene, camera);
```
michaelthatsit, line 449: This is what's causing the hit. I tried commenting it out and got the FPS back, and masking still worked.

hanbollar:
> > that's partially expected since ...
> 
> hmm but it doesn't look like you're applying that every update?
> 
> it's also not clear if it has anything to do with this PR. I went through and disabled each system to see where the perf hits might be coming from. disabling masking didn't change it.

🤔 odd - if this pr is creating any perf hits it should be coming from the addition of more passes in the render loop itself since it needs more than one this.renderer.render renderpass now (lotta render names haha) - https://github.com/Volumetrics-io/mrjs/pull/192/files#diff-9a9b14ca4cde264f45d74bc2e9e2926a929c81d6fccf4161133d5fc99d48d3faR441-R457

if this isnt causing the slowdown i can track down which pr it came from and use that as an addition to https://github.com/Volumetrics-io/mrjs/issues/205 - lmk ~

src/core/MRApp.js
hanbollar: sounds good ~

hanbollar:
wop i pressed enter before reloading - sounds good, looks like we're on the same page then - i'll clean it up per https://github.com/Volumetrics-io/mrjs/pull/192#issuecomment-1876097166 ~

src/core/MRApp.js
hanbollar: looks like threejs is comin in clutch ~

michaelthatsit (approved):
Wahoo! looks good, perf is back to where it was.

PR #211, opened 2024-01-04
optimization - only run system update in loop if necessary

hanbollar:


hanbollar:
putting a pause on this - other tasks are higher priority

hanbollar:
back to focus

PR #220, opened 2024-01-08
allow glb files to load with animation

hanbollar:


michaelthatsit:
Looks good! approved!

PR #232, opened 2024-01-11
optimization - only apply clipping on objects where stencil/masking not used

hanbollar:


hanbollar:
pausing this due to https://github.com/Volumetrics-io/mrjs/issues/233 i found in main (dont want to add more visual confusion with this pr while debugging that) - otherwise this pr is ready

hanbollar:
merging in since not a blocker

Issue #227, opened 2024-01-11
When dynamically creating an `mr-model`, you can't directly set the src property


lobau:
In this code, the assigning the `src` property directly doesn't work:

```js
let model = document.createElement("mr-model");

// that doesn't work
model.src = "./assets/fish.glb";

// needs setAttribute for it to work
model.setAttribute("src", "./assets/fish.glb");
```

PR #234, opened 2024-01-12
update to allow for `model.src` to update the attribute as expected

hanbollar:
Fixes https://github.com/Volumetrics-io/mrjs/issues/227


michaelthatsit (approved):
looks good!

PR #235, opened 2024-01-12
215 refactor anchoring api

michaelthatsit:
a lot happened here.

- added `mrjsUtils.xr` for direct access to the webXR API
- created AnchorSystem and `data-comp-anchor` to replace the SurfaceSystem and `mr-surface`
- created MRPlaneManaer to create and manage mr.js representation of detected planes.
- added collision groups and filtering to avoid raycasting against non-UI elements
- created draft interactive Anchors documentation
- shrank skybox and reduced fulstrum distance, should help with perf a bit.
- added better milkyway image, official NASA deep sky survey.


src/core/componentSystems/TextSystem.js
```diff
@@ -87,6 +88,11 @@ export class TextSystem extends MRSystem {
             }
 
             this.updateStyle(entity);
+            if (entity instanceof Button) {
```
hanbollar, line 91: just checking - is button the only current divEntity that needs a special case? or is it all div entities that can be within text

src/core/componentSystems/MRPlaneManager.js
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @function
+     * @description initializes the physics body of an MR Plane
+     * @param plane
+     */
+    initPhysicsBody(plane) {
+        const rigidBodyDesc = mrjsUtils.Physics.RAPIER.RigidBodyDesc.kinematicPositionBased().setTranslation(...this.tempPosition);
+        let colliderDesc = mrjsUtils.Physics.RAPIER.ColliderDesc.cuboid(...this.tempDimensions);
+        console.log(mrjsUtils.Physics.CollisionGroups.PLANES);
```
hanbollar, line 111: is this console.log still needed?

src/utils/XR.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @namespace xr
+ * @description Useful namespace for helping with xr utility functions.
+ * this is set within the MRApp to access various WebXR API features
+ */
+let xr = {};
```
hanbollar, line 6: Can we use `XR` instead of `xr` ?

src/core/entities/SkyBox.js
```diff
@@ -60,7 +61,7 @@ export class SkyBox extends MREntity {
             textureNames.map((name) => (path ? path + name : name)),
             this.onTextureLoaded.bind(this) // Ensuring the correct context
         );
-        const geometry = new THREE.SphereGeometry(1000, 32, 16);
+        const geometry = new THREE.SphereGeometry(900, 32, 16);
```
hanbollar, line 64: just curious on the change to 900, was it a render issue?

src/core/componentSystems/MRPlaneManager.js
```diff
@@ -0,0 +1,122 @@
+import { mrjsUtils } from 'mrjs';
+import { MRPlane } from '../../datatypes/MRPlane';
+
+/**
+ * @class MRPlaneManager
+ * @classdesc creates and manages the mr.js representation of XR planes.
+ * The resulting planes have RAPIER rigid bodies and THREE.js meshes that occlude virtual content by default
+ */
+export class MRPlaneManager {
```
hanbollar, line 9: if this is a manager and not a system we need to figure somewhere else for this to go - to default back to keeping inline with the ECS setup we're going for: one option to actually just change this into a System and just have an empty update function (like how the `MaskingSystem` is setup)

samples/assets/textures/starmap_4k.jpg
hanbollar, line 1: if this is a skybox - can you pre-pend its name with `skybox_` just to have easier discoverability by users?

hanbollar (approved):
a quick go through - lmk if any questions - nothing major ~

src/core/componentSystems/TextSystem.js
michaelthatsit: only button.

src/core/componentSystems/MRPlaneManager.js
michaelthatsit: It's not, good catch.

src/utils/XR.js
michaelthatsit: xr is consistent with the web standard `navigator.xr` or `renderer.xr`

src/core/entities/SkyBox.js
michaelthatsit: I shrank down the fulstrum distance from 10000 to 1000, which put this right up on the edge, causing clipping when you walk around.

src/core/componentSystems/MRPlaneManager.js
michaelthatsit: Ah yes I was thinking that but forgot to move it. I'd prefer to only use systems when they're actually taking advantage of the architecture, otherwise it's a wasted update call.

Issue #205, opened 2024-01-02
seems like there's a memory leak somewhere...


michaelthatsit:
Not quite sure where it might be, but if you leave the demo running long enough on headset, we start getting WebGL warnings and some of the panels begin to flicker.

michaelthatsit:
@hanbollar added you too it. I'll get a recording of it later.

michaelthatsit:
@hanbollar I'm running into this something terrible with the anchoring sample. I'm 90% certain it's that we're destroying/creating geometries for UI elements. But it could also be text rendering. We should prioritized investigating this.

hanbollar:
> @hanbollar I'm running into this something terrible with the anchoring sample. I'm 90% certain it's that we're destroying/creating geometries for UI elements. But it could also be text rendering. We should prioritized investigating this.

sg, i'll make that a top priority for me for today ~

michaelthatsit:
<img width="1493" alt="Screenshot 2024-01-11 at 3 32 46 PM" src="https://github.com/Volumetrics-io/mrjs/assets/19628643/57d9c4b4-af15-4251-8c98-655d4fffe199">

captured when things started freaking out. I have a video as well, waiting for my quest to finish charging so I can get it on my mac.

but the simple description is that the UIPlanes begin to flicker as the warnings start rolling in, eventually the session restarts.

hanbollar:
[...]

PR #236, opened 2024-01-12
optimization - clean up system and entity style updates

hanbollar:
related to https://github.com/Volumetrics-io/mrjs/issues/205


michaelthatsit:
A heads up I made an update in this and text system which gave us a boost in performance, I merged it so you'll need to merge in the changes from main.

hanbollar:
given the overlap in naming and use - im going to rethink the needs style update functions for better integration with the flag and vice versa - going to pause and noodle on this a short bit

hanbollar:
@michaelthatsit - short explanation:
- simplifies the update check: changes out the old `needsUpdate` for the system calls to `needsSystemUpdate` and creates a `needsStyleUpdate` for entity items
- follows the `_needsStyleUpdate` and `_needsSystemUpdate` flags in `MREntity` and `MRSystem` respectively where the user only needs to use the getters and setters to have them run as expected - using the trifecta of true/false allows triggering and takes prescedent and setting to undefined allows it to run every call until some outside interference sets it back to true/false.

tldr - first check out `MREntity` and `MRSystem` changes, then the changes in the other files will make sense

also added an MRImage fix into the pr since the async loading problem became more apparent with this setup ~

src/core/componentSystems/StyleSystem.js
```diff
@@ -14,13 +14,34 @@ export class StyleSystem extends MRSystem {
...
 
+    /**
+     * @function
+     * @description Getter to checks if we need to run this system's update call. Overridden implementation returns true if there are any items in this
+     * systems registry that need to be run AND the default systemUpdateCheck is true
+     * (see [MRSystem.needsSystemUpdate](https://docs.mrjs.io/javascript-api/#mrsystem.needssystemupdate) for default).
+     * @returns {boolean} true if the system is in a state where this system is needed to update, false otherwise
+     */
+    get needsSystemUpdate() {
+        return this.registry.size > 0 && super.needsSystemUpdate;
```
michaelthatsit, line 34: this needs to be always true and we need to stick with checking `entity.needsStyleUpdate` due to some asynchronous loading of text and event driven styles such as hover and click. `needsSystemUpdate` blocks these.

src/core/MRSystem.js
```diff
@@ -71,10 +93,13 @@ export class MRSystem {
      * @param {number} deltaTime - given timestep to be used for any feature changes
      * @param {object} frame - given frame information to be used for any feature changes
      */
-    __update(deltaTime, frame) {
-        if (!this.needsUpdate(deltaTime, frame)) {
+    _update(deltaTime, frame) {
+        console.log('inside MRSystem update function');
```
michaelthatsit, line 97: remove log

src/core/MRSystem.js
```diff
@@ -71,10 +93,13 @@ export class MRSystem {
...
+        if (!this.needsSystemUpdate) {
+            console.log('not updating');
```
michaelthatsit, line 99: remove log

michaelthatsit:
@hanbollar minor clean up and styleSystem needs to rely on check entity.needsStyleUpdate, with `needsSystemUpdate` set to always true. it should have the same effect given that most of the time an entity won't need a style update.

src/core/componentSystems/StyleSystem.js
hanbollar: perf - yea i wasnt sure about this one specifically, good to know ~

michaelthatsit: Screenshot 2024-01-22 at 10 59 49 AM

example of the issue it causes, text loads on a delay so it's ancestors need to be told to update once it's ready.

hanbollar: wait - nvm - that's what it's doing i thought - one sec

hanbollar: i think this part specifically is what's causing the delayed trigger then:

https://github.com/Volumetrics-io/mrjs/pull/236/files/260bcdc7ad84140e87906599d29b53ada2d6da7e#diff-c755f975e32bd8242038b2c9e13c3c9c01b48119dee283b4898964e3b7b8431eR18-R22

because otherwise this.needsSystemUpdate stays undefined and will always run

michaelthatsit: > going to enforce it 100% for this one - but then what is the point of 
> 
> ```
>        // want this system to run based on the true/false trigger
>         this.needsSystemUpdate = true;
> 
>         document.addEventListener('panel-mutated', () => {
>             this.needsSystemUpdate = true;
>         });
>         ```

Ah this probably isn't necessary anymore.

hanbollar: gotcha - cleaning up, and letting it revert back to just the super's implementation then ~

PR #237, opened 2024-01-12
add needsUpdate for AnchorSystem

hanbollar:
heads up @michaelthatsit - forgot to note this on the pr, adding it in since it's a small fix to the merged version - it's a change based on the optimization from https://github.com/Volumetrics-io/mrjs/pull/211

im going to merge this to get it going, but just wanted to give u a heads up

### DIFF
--- a/src/core/componentSystems/AnchorSystem.js
+++ b/src/core/componentSystems/AnchorSystem.js
@@ -107,6 +107,19 @@ export class AnchorSystem extends MRSystem {
 
     /**
      * @function
+     * @description Checks if we need to run the generic system update call. Default implementation returns true if there are
+     * any items in the system's registry. Allows subclasses to override with their own implementation.
+     * @param {number} deltaTime - given timestep to be used for any feature changes
+     * @param {object} frame - given frame information to be used for any feature changes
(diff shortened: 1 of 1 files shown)


michaelthatsit:
Wait how are we indicating when something needs an update? Anchors can move anytime so its one of those things that needs to be updated every frame to keep transforms in sync with the real world anchor.

hanbollar:
> Wait how are we indicating when something needs an update? Anchors can move anytime so its one of those things that needs to be updated every frame to keep transforms in sync with the real world anchor.

yea this is to make sure the needsUpdate always runs as true - otherwise it'll rely on if the registry is set or not per: https://github.com/Volumetrics-io/mrjs/blob/main/src/core/MRSystem.js#L75 , so adding in this change lets it still run every frame (ie the return true)

lmk if this makes sense - can hop on a quick call to explain as well if that helps

hanbollar:
updated the comment in the function for better clarity as well - https://github.com/Volumetrics-io/mrjs/pull/238

michaelthatsit:
Ah ok. I'm out rn but I'll take a closer look when I get back. But sounds good to me.

PR #250, opened 2024-01-16
add geometry disposal for Image with UIPlane before resetting

hanbollar:


hanbollar:
separating out from https://github.com/Volumetrics-io/mrjs/pull/236

PR #251, opened 2024-01-17
Revert "optimization - only apply clipping on objects where stencil/masking not used"

hanbollar:
Reverts Volumetrics-io/mrjs#232


hanbollar:
brings back mrimage - todo to look into before allowing this pr back in

PR #253, opened 2024-01-17
update naming for better `MR` structure

hanbollar:


hanbollar:
heads up @michaelthatsit - moved the datatypes and datamanagers out of core for better separation, updated the file names for entities and their classes, and cleaned up some includes

lmk if any questions

michaelthatsit:
Looks good, but img still isn't loading properly on chrome/arc for some reason, it's fine in headset. I'm also running locally so I don't think it's a render thing. I can investigate separately.

michaelthatsit (approved):

PR #259, opened 2024-01-18
added event to fix responsiveness bug, also temp fix for text masking…

michaelthatsit:
… issue and clipping bugfix we didn't catch.

### DIFF
--- a/src/core/componentSystems/ClippingSystem.js
+++ b/src/core/componentSystems/ClippingSystem.js
@@ -89,7 +89,6 @@ export class ClippingSystem extends MRSystem {
      * @param {MREntity} entity - the entity to which we're adding the clipping planes information
      */
     addClippingPlanes(entity) {
-        console.log(entity);
         this.geometry = entity.clipping.geometry.toNonIndexed();
         let geoPositionArray = this.geometry.attributes.position.array;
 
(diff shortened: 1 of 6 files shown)


michaelthatsit:
@hanbollar while fixing the responsiveness bug I caught an issue with clipping/masking text.  for `MRText` and it's siblings, the text mesh is in `textObj` and not `object3D`, this is admittedly not ideal and I'm still noodling on how to eliminate this. BUT it also means that the masking stencil isn't being applied.

We're probably just catching this because up until now, it was being clipped instead of masked. But there was a bug we missed in your recent update where the clipping planes weren't being created/applied.

I'll leave this up til tomorrow so you can take a look but we should merge this until we find a better solution, since main is technically broken.

src/core/componentSystems/MaskingSystem.js
```diff
@@ -4,6 +4,7 @@ import { MRSystem } from 'mrjs/core/MRSystem';
 import { MRDivEntity } from 'mrjs/core/MRDivEntity';
 import { MREntity } from 'mrjs/core/MREntity';
 import { MRPanel } from 'mrjs/core/entities/MRPanel';
+import { MRTextEntity } from '../MRTextEntity';
```
hanbollar, line 7: change this to `import { MRTextEntity } from 'mrjs/core/MRTextEntity';`

src/core/componentSystems/ClippingSystem.js
```diff
@@ -89,6 +89,7 @@ export class ClippingSystem extends MRSystem {
      * @param {MREntity} entity - the entity to which we're adding the clipping planes information
      */
     addClippingPlanes(entity) {
+        console.log(entity);
```
hanbollar, line 92: if this is for just-this-pr debugging dont forget to remove

PR #276, opened 2024-01-22
115 add an audio parameter to mr entity so any entity can emit sound

michaelthatsit:


hanbollar (approved):
lgtm - good to go ~

this file needs docs fill in : https://github.com/Volumetrics-io/mrjs/pull/276/files#diff-59f2f8a278182aea0418b7233cbbc5fbbaba033d3b4363caa5552f2a0196d1c4R1
 but if that's not a priority no worries

PR #298, opened 2024-01-24
297 plane anchors bug

michaelthatsit:


src/core/MRSystem.js
```diff
@@ -107,10 +107,12 @@ export class MRSystem {
...
-        // reset update var if needed.
-        if (!this.alwaysNeedsSystemUpdate) {
-            this.needsSystemUpdate = false;
-        }
+        // this should be revisited, the update there are situations where this 
+        // needs to remain true until a task has been done, specifically the anchoring system
+        // // reset update var if needed.
+        // if (!this.alwaysNeedsSystemUpdate) {
+        //     this.needsSystemUpdate = false;
+        // }
```
michaelthatsit, line 115: @hanbollar we should revisit this, it breaks the anchoring system. Which I don't want always updating but it needs to update until after plane anchors have been set. we can discuss more tomorrow.

PR #306, opened 2024-01-24
models, skybox examples && text traverse fix

hanbollar:


hanbollar:
merging to get it in - adding more details to 'enhancements' issue for skybox - https://github.com/Volumetrics-io/mrjs/issues/319

src/core/MRTextEntity.js
```diff
@@ -45,6 +45,23 @@ export class MRTextEntity extends MRDivEntity {
...
     }
+
+    /**
+     * @function
+     * @description Runs the passed through function on this object and every child of this object.
+     * @param {Function} callBack - the function to run recursively.
+     */
+    traverse(callBack) {
+        callBack(this);
+        const children = Array.from(this.object3D.children);
```
takahirox, line 56: Is this `this.object3D.children` (Three.js Object3D collection) intentional although parent `MREntity` traverses `.children` (HTMLElement collection)? I guess it isn't. This function it very trouble some because callback function doesn't expect `Object3D` argument if this method is called from parent entities.

PR #309, opened 2024-01-25
302 create mr volume element

michaelthatsit:
reimplemented `mr-volume` for clipping anchored content, even when it's not UI.

also made improvements to water shader and fixed an anchoring bug.


src/core/MREntity.js
```diff
@@ -65,6 +65,8 @@ export class MREntity extends MRElement {
         this.touch = false;
         this.grabbed = false;
         this.focus = false;
+
+        this.ignoreStencil = true
```
hanbollar, line 69: question - this should make all entities ignored by masking, even div entities - meaning we're not masking at all anymore - is this needed in this pr?

if so - make sure to add `this.ignoreStencil = false` to all mrdiventity items (or at a basis, mrdiventity)

hanbollar (approved):
one comment that needs looking at before merge - once that's resolved / lmk if you have questions - then lgtm ~

src/core/MREntity.js
michaelthatsit: ah it should be disabled for MREntity but enabled for MRDivEntity, I'll fix that. This should achieve the desired effect of masking UI elements but allowing 3D content to be clipped in an mr-volume.

michaelthatsit: @hanbollar looks like this is already the case in MRDivEntity, which sets it's values after the MREntities constructor runs.

https://github.com/Volumetrics-io/mrjs/blob/c802e57ef5f70f8f0ee3f25f915844c0550e745f/src/core/MRDivEntity.js#L44

hanbollar: ah shoot - nah you're all good there, just checked and MRDivEntity already has the this.ignoreStencil = false ~, good to merge

PR #311, opened 2024-01-25
Fix for text updating based on changed `innerText` value of an `mr-text`

hanbollar:


src/core/componentSystems/TextSystem.js
```diff
@@ -81,7 +75,11 @@ export class TextSystem extends MRSystem {
      * Relies on the parent's implementation. (see [MRSystem.needsSystemUpdate](https://docs.mrjs.io/javascript-api/#mrsystem.needssystemupdate) for default).
      */
     set needsSystemUpdate(bool) {
-        super.needsSystemUpdate = bool;
+        console.log('Updating a value that does not need updating. TextSystem`s `needsSystemUpdate` solely depends on registry size \
+            instead of a boolean. This is required s.t. we can monitor and properly update the system when the text content has changed \
```
michaelthatsit, line 79: is this log necessary?

hanbollar: yes and no - yes to inform the user that this function technically does nothing though the function is needed to be implemented - no in that i can make the log a lot shorter and keep most of it as a code comment.

i can switch it to the code comment + small message version - one sec

michaelthatsit (approved):

PR #333, opened 2024-01-28
325 style system update bug

michaelthatsit:


michaelthatsit:
@hanbollar there are some changes I had to revert to fix the perf and crash bug, let's chat about it monday but I'm gonna merge this. Don't want folks getting hit with crashing browsers the first time they try out the library.

hanbollar:
> @hanbollar there are some changes I had to revert to fix the perf and crash bug, let's chat about it monday but I'm gonna merge this. Don't want folks getting hit with crashing browsers the first time they try out the library.

@michaelthatsit oh no worries - glad you got the fix in

looks like no more changes are needed though since it seems your fix removed some of the mixed changes from when the style system was still IP

i think what we have now from these changes can count as a final version unless there's a few parts you'd still want me to look at to optimize for it

PR #348, opened 2024-02-02
fixed scale bug.

michaelthatsit:


michaelthatsit:
@hanbollar I added a global directly to mrjsUtils so lemme know if you think there'd be a better place for it, I also have other globals we could probably coral under mrjsUtils or maybe create mrjsEnums?

@adnanwahab there's a slight change that will impact the TextSystem so you'll wanna merge in my changes when this hits main.

src/utils/index.js
```diff
@@ -30,6 +30,13 @@ const mrjsUtils = {
     Physics,
     StringUtils,
     xr,
+
+    get appScale() {
```
hanbollar, line 34: this should be a function within mr-app even if it is not object dependent - there should be a way to do a `static` function of the class in javascript for it

hanbollar (approved):
> I added a global directly to mrjsUtils so lemme know if you think there'd be a better place for it, I also have other globals we could probably coral under mrjsUtils or maybe create mrjsEnums

@michaelthatsit for the above - i think it would be useful for us to have one `enums.js` at the same level as `index.js` and `globals.js` at the root of the project, i dont think having it as a completely separate subproject like mrjsUtils is useful

-- also see my other comment about the one function you had added to index.js

otherwise lgtm 🥳

src/utils/index.js
michaelthatsit: I thought about this but not everything has access to `mr-app`. and I'd like to keep it that way, as it'd make things harrier.

hanbollar: gotcha, then doing something in like `utils/app.js` should be fine (or whatever js you want to call it in `utils/...`) - we just shouldnt directly put a function in the index.js is all for better scalability and future testing

michaelthatsit: Yeah that's a fair point. I'm cooking and haven't tested your PR yet. I'll make the change after finishing both of those.

hanbollar: yea no rush ~

PR #381, opened 2024-02-08
fix layout for odd spacing

hanbollar:


hanbollar:
doesnt fix it by itself --> this cant be separate from the style system fix as it's due to a few lingering issues + this then

PR #395, opened 2024-02-11
a small fix for the geometry update

hanbollar:
temp solution before we discuss more about the needsStyleUpdate && needsGeometryUpdate

has a fix for the bubbling

i like the idea of doing the checks with the returns and possibly just no longer needing the `needs...Update` functions at all

-----

but i also like the idea in future if we add a 'predetermine' system that goes through all the system entities beforehand and determines what actually needs to update and which functions within the update for that system that it actually needs to remove the use of so many if/else branchings within the style/geometry/other systems

too high level for right now but it's a noodling thought


hanbollar:
since this is a small fix and we'll be going over a better cleanup starting next week - just going to get this in

cleanup/refactor later will simplify this / mrapp / other systems more cleanly

hanbollar:
after prs like this we can probably remove the entire need for `needsUpdate` type of checking, but that's something to be discussed next week - not priority

PR #396, opened 2024-02-11
Allow for opacity css values in material background color

hanbollar:
- allow for color with opacity

----

~~note: has some changes from the #395 pr but that's just bc this was building off of it~~

---

updated comment here: this is an old pr that just needed to sit since we were still changing the updates for the `MaterialStyleSystem` - now that that's in place, this just adds an additional coloring feature to the system


hanbollar:
pausing this in favor of image pr

hanbollar:
moving this out of draft mode since it's ip again

hanbollar:
marking this as paused

hanbollar:
working on this again

hanbollar:
Cleaned up - main change introduced by this old pr is the updated to the `setBackgroundColor` function in the `MaterialStyleSystem`

good quick way to test it is to change the background color of navbar in samples/index.html

allows for: css color, rgb, rgba, hex, hex with alpha

michaelthatsit (approved):
works! assuming you tested in headset.

PR #403, opened 2024-02-13
389 xr button session fix

michaelthatsit:
- added scene origin solution
- added default floor clipping plane
- fixed button z-index default


src/core/componentSystems/LayoutSystem.js
```diff
@@ -102,7 +102,7 @@ export class LayoutSystem extends MRSystem {
 
         if (entity.compStyle.zIndex != 'auto') {
             // default zIndex values in css are in the 1000s - using this arbitrary divide to convert to an actual usable threejs value.
-            entity.object3D.position.setZ(parseFloat(entity.compStyle.zIndex / 1000));
+            entity.object3D.position.setZ(parseFloat(entity.compStyle.zIndex) / 1000);
```
hanbollar, line 105: do we also need this fix in the panel system?

hanbollar (approved):
approved - one question related to the change in layout system for panel system since i think they had the same code there (so they might hv the same bug fix needed)

otherwise lgtm

PR #431, opened 2024-02-21
fix image-styling, `object-fit`: `contain`, `fill`, `cover`, `scale-down`, `none`

### LINKED ISSUE #430: make all image (obj vs texture) setups work as expected
combining the below into one issue since theyre related
- there's some texture vs object discrepancies depending on cover/contain/fill/...
- there's also a width issue with panel at the base - the below image demos width being fine, but not fine for panel (and all items that are inside it are fine, theyre just shifted given the hierarchy)
Screenshot 2024-02-20 at 8 06 02 PM

starting becoming a noticeable problem from [365](https://github.com/Volumetrics-io/mrjs/commit/520a6a021e036bb86fcaafc345853217b6f67935) but mightve been a hidden issue prior to that commit so more math to dig through

hanbollar:
Fixes #430


hanbollar:
dug through width styling difference - related to overflow - see [link](https://github.com/Volumetrics-io/mrjs/issues/430#issuecomment-1958004208) for the rest of the commentary - that's a styling issue not an actual mrjs setup issue - tabling that instead of as part of this pr

hanbollar:
contain now works in separate temp impl - now to pass that back to the mrjs setup

hanbollar:
> dug through width styling difference - related to overflow - see [link](https://github.com/Volumetrics-io/mrjs/issues/430#issuecomment-1958004208) for the rest of the commentary - that's a styling issue not an actual mrjs setup issue - tabling that instead of as part of this pr

this part is all good now per https://github.com/Volumetrics-io/mrjs/issues/439#issuecomment-1965236031

hanbollar:
> contain now works, though it has a lingering visual effect that gets cleared on page reload - looking into:
> 
> before reload (after squishing ...

contain and scale-down are happy now - no more windows xp look

doing some last cleanup then will set for review

hanbollar:
@michaelthatsit ready for review - quick way to test the options is to change the value of the `object-fit` [here](https://github.com/Volumetrics-io/mrjs/pull/431/files#diff-b9a7239b3568efaf325832b0b99b18ad111ef950451d74be20c1e9052ca4e8b6R237)

michaelthatsit:
image

michaelthatsit:
It looks good to me. Dynamically changing between each value resulted in weird behavior but this can land and we can fix that separately.

PR #436, opened 2024-02-21
Deamoner/issue363

Deamoner:
Ok mostly got mr-video working and it has a screenshare demo built into the example. Few notes:
- No mutation triggered to change size when entering XR mode - seems like a base change to the base div entity but I could be off there
- Someting happens when exiting XR
Video of demo:
https://app.screencastify.com/v3/watch/DhvI51L9SAVS4j21fZO1


samples/examples/video.html
```diff
@@ -0,0 +1,120 @@
...
+    <link rel="stylesheet" type="text/css" href="video-style.css" />
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body >
+   <mr-app>
+    <mr-entity id="rig" data-comp-anchor="type: fixed;">
+        <mr-panel id="tabbar" data-position="0 0 0.001"> <!--TODO: move Z offset to css z-index-->
+            <mr-button onclick="playVideo()" class="col-1">Play</mr-button>
+            <mr-button onclick="stopVideo()" class="col-2">Stop</mr-button>
+            <mr-button onclick="screenShare()" class="col-3">Screenshare</mr-button>
```
michaelthatsit, line 17: This is really cool! but probably shouldn't make it into the simple example. Also it doesn't work when in headset. Maybe we replace it with a loop button?

src/core/entities/MRVideo.js
```diff
@@ -0,0 +1,211 @@
...
+import { MRDivEntity } from 'mrjs/core/MRDivEntity';
+
+import { mrjsUtils } from 'mrjs';
+
+/**
+ * @class MRVideo
+ * @classdesc Base html video represented in 3D space. `mr-video`
+ * @augments MRDivEntity
+ */
+export class MRVideo extends MRDivEntity {
```
michaelthatsit, line 12: @hanbollar you're currently working on the fix for `object-fit`, MRVideo should be treated the same as MRImage. Perhaps we can extract their common features out into MRMedia? that wouldn't need to be exposed as a tag itself, but might make it easy for us to apply the same responsiveness logic to both.

michaelthatsit (changes requested):

src/core/entities/MRVideo.js
hanbollar: yea we can def do something like that the same way we're doing mrtext --> mrtextfield && mrtextarea

i'll make an open todo to do that after the mrimage fix 👍

hanbollar: noting: #437

hanbollar (approved):

hanbollar:
thanks for helping out with these changes @Deamoner !!

PR #447, opened 2024-02-26
Spotlight occlusion

michaelthatsit:
This is semi-immersive experiences such as the skybox demo. users can still see there body and part of their environment while being immersed and another environment.


hanbollar (approved):
if `MRUser` is being placed in core, then `MRHand` should as well,

otherwise MRUser should be placed in dataTypes

---

i'm of the preference to put both in core though

--------

otherwise lgtm 🥳

michaelthatsit:
Moved both to their `core/user/` as there might be more features we'll wanna add later.

PR #456, opened 2024-02-27
fix for npm publishing action

hanbollar:
moving into a pr because this quick fix is starting to spam main


hanbollar:
omg it works, merging

Issue #454, opened 2024-02-27
dynamically changing between `object-fit` values results weird behavior


hanbollar:
It looks good to me. Dynamically changing between each value resulted in weird behavior but this can land and we can fix that separately.

_Originally posted by @michaelthatsit in https://github.com/Volumetrics-io/mrjs/issues/431#issuecomment-1967517801_

PR #464, opened 2024-02-28
add MRMedia base for MRImage and MRVideo

hanbollar:
(this also includes an `npm run prettier-fix` so there's some additional small changes separate from the main pr-focus)


hanbollar:
this seems happen for image - looking through it for video now too

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        this.style.width = `${this.objectFitDimensions.width}px`;
+        this.style.height = `${this.objectFitDimensions.height}px`;
+    }
+}
+
+// TODO - dont want to allow users to create this as a generic item, just as a base class for all future
+// media elements?
+// customElements.get('mr-media') || customElements.define('mr-media', MRMedia);
```
hanbollar, line 242: thoughts?

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+    mutated(mutation) {
+        super.mutated();
+
+        // moving the if 'mutation' handling check to the children, since
+        // mutations are only understood by their actual type. Any mutation
+        // passed through MRMedia directly is undefined since it is not
+        // a direct element for users.
+        //
+        // those functions can do the if (mutation.type ....) handling and
+        // their specific cases and then call back up to here to run the below functionality.
```
hanbollar, line 120: thoughts?

i dont really like doing it this way but it doesnt work as a pure super call from the children because of how 'mutation' is passed into the function based on entity type -- but also didnt want to reimplement the mutated for every child so this was a decent go-between

michaelthatsit: super is a good solution!

michaelthatsit:
combing through it now but obligatory question: 

Did you test in headset?

hanbollar:
> combing through it now but obligatory question: 
> 
> 
> 
> Did you test in headset?

yes 🤓

michaelthatsit:
Looks good but found two bugs. 

1. `contain` seems to be broken in XR. At least with video, I didn't check against image. 

2. When exiting XR it doesn't refit to the window. I'll try to grab a debug log

Uploading videos for both issues

michaelthatsit:
contain bug

https://github.com/Volumetrics-io/mrjs/assets/19628643/0fab2abb-0098-4796-b93d-7887fd16557b

michaelthatsit:
refit bug

https://github.com/Volumetrics-io/mrjs/assets/19628643/5ebf2cbb-3c02-468b-a520-f5241420fd0b

michaelthatsit:
Last note, once those are fixed, let's make `contain` the default property for both images and videos in defaultStyle

src/utils/Material.js
```diff
@@ -82,13 +82,39 @@ material.loadTextureAsync = function (src) {
...
-        try {
-            const textureLoader = new THREE.VideoTexture(video);
-            resolve(textureLoader);
-        } catch(err) {
-            reject(err);
-        }
+        // Event listener to ensure video is ready
+        video.onloadeddata = () => {
+            try {
+                const textureLoader = new THREE.VideoTexture(video);
```
takahirox, line 94: Nit for an existing issue: Isn't the name `textureLoader` for `VideoTexture` instance confusing because Three.js has `TextureLoader` class. Maybe `texture` or `videoTexture` would be clearer?

src/core/componentSystems/MaterialStyleSystem.js
```diff
@@ -36,6 +36,9 @@ export class MaterialStyleSystem extends MRSystem {
             if (entity instanceof MRDivEntity) {
                 this.setBackground(entity);
             }
+            if (entity instanceof MRVideo && entity.playing) {
+                entity.texture.needsUpdate = true;
+            }
```
takahirox, line 41: Is a texture of `MRVideo` guaranteed to be as `VideoTexture`? If so we may not need to manually call `texture.needsUpdate = true` because it's automatically (and more efficiently with [requestVideoFrameCallback](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestVideoFrameCallback))

https://threejs.org/docs/#api/en/textures/VideoTexture.needsUpdate

Even if it isn't guaranteed, we may add a guard like

```
if (entity instanceof MRVideo && entity.playing && !entity.texture.isVideoTexture) {
  entity.texture.needsUpdate = true;
}
```

samples/examples/video.html
```diff
@@ -37,81 +28,74 @@
...
+        let panels = document.querySelectorAll('.layout');
+        let videoEl = document.querySelector('#videoEl');
 
-        let frontIndex = 0
+        let frontIndex = 0;
 
         function playVideo() {
-            let videoEl = document.querySelector('#videoEl')
-            videoEl.play()
+            videoEl.play();;
```
takahirox, line 38: Nit: Duplicated ";"

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+     * @class
+     * @description Constructs a base media entity using a UIPlane and other 3D elements as necessary.
+     */
+    constructor() {
+        super();
+        this.attachShadow({ mode: 'open' });
+
+        // Create the object3D. Dont need default value for geometry
+        // until the connected call since this will get overwritten anyways.
+        let material = new THREE.MeshStandardMaterial({
```
takahirox, line 23: I would like to always use `const` specification instead of `let` for immutable variables to avoid unexpected reassignment.

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+            side: 0,
```
takahirox, line 24: Using named Three.js constants, `THREE.FrontSide` in this case, would be more readable

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+            transparent: true,
+        });
+        // Object3D for MRMedia (mrimage,mrvideo,etc) is the actual image/video/etc itself in 3D space
+        this.object3D = new THREE.Mesh(undefined, material);
+        this.object3D.receiveShadow = true;
+        this.object3D.renderOrder = 3;
```
takahirox, line 30: Not a review comment but just a question for existing code so not blocking: What is this `renderOrder = 3` for? Adding a comment would be helpful to users.

src/core/MRMedia.js
```diff
@@ -0,0 +1,242 @@
...
+     * @description Calculates the height of the img based on the img tag in the shadow root
+     * @returns {number} - the resolved height
+     */
+    get height() {
+        let height = this.objectFitDimensions?.height;
+        return height > 0 ? height : super.height;
+    }
+
+    get mediaWidth() {
+        // to be filled in in children
```
takahirox, line 68: Nit, not blocking issue: Do we want to throw an error here? We may be able to more easily detect in cases if we forgot to override this method in child classes.

src/utils/Material.js
```diff
@@ -82,13 +82,39 @@ material.loadTextureAsync = function (src) {
...
+                    .play()
+                    .then(() => {
+                        console.log('Video playback started');
+                    })
+                    .catch((e) => {
+                        console.error('Error trying to play the video:', e);
+                        reject(e);
+                    });
+                resolve(textureLoader);
+            } catch (err) {
```
takahirox, line 107: Curious, which line above can throw an un-caught error? video play error stuffs seems to be caught above.

hanbollar: it's to catch the line associated with `const videoTexture = new THREE.VideoTexture(video);` where if the browser handling is not setup right based on context or autoplay of video, it'll error at the VideoTexture creation

i think the other changes ive added make it so there should be no other errors that actually hit this, but given browsers' potential to change what can autoload as a webpage loads, i thought this would be useful - it's more of a failsafe for a failsafe at the moment - but i can remove it if you think that wont ever be necessary - let me know ~

src/core/MRMedia.js
hanbollar: √

src/core/MRMedia.js
hanbollar: √ and added it to other places where we manually created side as well ~

src/core/MRMedia.js
hanbollar: √

samples/examples/video.html
hanbollar: √

src/core/componentSystems/MaterialStyleSystem.js
hanbollar: good to know, since it's guaranteed, i'll remove it ~

hanbollar: √

src/utils/Material.js
hanbollar: √

hanbollar:
> contain bug
> 
>  cannot-be-contained.mp4

odd and good catch 🤔  - not sure why that didnt show up on my end, i wonder if it was a reloading thing based on how i was serving to the headset?

i know what's causing this so this should be cleared up in the next commit pass i do

hanbollar:
given the additional passes from today, i'll also add the fix for https://github.com/Volumetrics-io/mrjs/issues/454 in this pr then as well

src/core/MRMedia.js
hanbollar: right now my understanding for mrjs is we're using renderOrder kind of like an addition to Z-depth to get around some of the overhead of that, and then i think we were going to switch it to an enum to make it cleaner-code so a comment wouldnt be necessary

i'm not fully up to speed on it otherwise though, but @michaelthatsit would have a better answer to add here if i missed anything in my above explanation

src/core/MRMedia.js
hanbollar: went through this again and looks i /can/ actually just remove the children's functions - must have erroring due to some of the ip work i had while i was setting this up - cleaned up and just using the parent version now ~

src/core/MRMedia.js
hanbollar: removing ~

hanbollar:
(while i finish cleaning this up pulled back the review to not spam the notifications since commentary was already supplied)

hanbollar:
> > refit bug
> > refit-trim.mp4
> 
> @michaelthatsit i'm a bit confused for this bug you noted - i'm not fully seeing an issue in the image comparing the before and after of XR for that panel - can you grab a still from the video-clip you made to point what you mean or explain it a bit more in words here?

@michaelthatsit are you specifically talking about the part where the page ends up white when exiting MR? or is there a different formatting thing you're noting

Screenshot 2024-03-01 at 1 34 20 PM

hanbollar:
did this separately to not overload this pr: https://github.com/Volumetrics-io/mrjs/pull/472

src/core/MRMedia.js
hanbollar: i think this older branch might've also been related to it: `349-integrate-depth-sensing-to-replace-render-order-hack`

hanbollar:
remaining todos:
- waiting on answers from some comments here
- visual check to confirm mr-fix
- make sure that the other issue was also fixed in the process

src/utils/Material.js
takahirox: > it's to catch the line associated with const `videoTexture = new THREE.VideoTexture(video);` where if the browser handling is not setup right based on context or autoplay of video, it'll error at the VideoTexture creation

If I'm right, I think you might be mistaken about this point because Three.js `VideoTexture` constructor doesn't include any video element specific code except for [these lines](https://github.com/mrdoob/three.js/blob/6e897f9a42d615403dfa812b45663149f2d2db3e/src/textures/VideoTexture.js#L19-L30).

So I don't think it can cause an error regardless of video status or configuration. It doesn't throw an error even with just an empty object.

```
new THREE.VideoTexture(document.createElement('video'));
-> VideoTexture {isTexture: true, uuid: ...}
new THREE.VideoTexture({})
-> VideoTexture {isTexture: true, uuid: ...}
```

I think unnecessary try-catch blocks should be avoided if possible.

src/core/MRMedia.js
takahirox: @michaelthatsit I'm happy if you explain because as I explained at other places I'm thinking of using `renderOrder` for masking (stencil) system.

Anyways, this is just a question, not a comment for any blocking issue.

michaelthatsit:
> > > refit bug
> > > refit-trim.mp4
> > 
> > 
> > @michaelthatsit i'm a bit confused for this bug you noted - i'm not fully seeing an issue in the image comparing the before and after of XR for that panel - can you grab a still from the video-clip you made to point what you mean or explain it a bit more in words here?
> 
> @michaelthatsit are you specifically talking about the part where the page ends up white when exiting MR? or is there a different formatting thing you're noting
> 
> Screenshot 2024-03-01 at 1 34 20 PM

Sorry for the delay.

The expectation is that the video be exactly where it was when first opening the browser, prior to entering XR. Instead we're greeted with a blank panel. The menu bar is visible at the bottom. so it's something to do with the `mr-video` specifically. Also worth noting this doesn't happen with `mr-image`.

src/utils/Material.js
hanbollar: sounds good, was just trying to be extra cautious - i'll make sure that try-catch block is removed :)

hanbollar:
remaining todo for me:
- test in headset to confirm is all good

src/core/MRMedia.js
hanbollar: marking this as resolved just to make sure i can keep an eye on remaining todos for this pr

opened up a [separate issue](https://github.com/Volumetrics-io/mrjs/issues/480) so @michaelthatsit can follow up with that discussion for you @takahirox :)

hanbollar:
headset 🥽 now shows a scaling issue after my fixes for the other linked issues

going to update my headset since it's been a bit and check again just in case

otherwise this is priority 1 for tomorrow

hanbollar:
@michaelthatsit - in headset all good now - thankfully the issue of the white had to do with the example itself so all gucci ~

final notes:
- defaultCSS has mr-video and mr-img with object-fit:contain
- our main example's index.html overwrites it to cover for visual purposes
- added an images example for object-fit comparisons long term as well

lmk if there's anything else or any final notes for the review

hanbollar:
rerequesting review from both, but up to y'all for what's needed ~

michaelthatsit:
@hanbollar Just tested and all works as expected. Thanks for pushing this through.

michaelthatsit (approved):

Issue #441, opened 2024-02-22
`mr-stats` element for tracking perf in XR


michaelthatsit:
Should include

- FPS
- memory
- latency
- ??

takahirox:
It may be the easiest solution to bring the one in [WebXR samples](https://immersive-web.github.io/webxr-samples/) to MRjs if possible

michaelthatsit:
I'd prefer to have something that looks more native to MRjs, but that should work for now. Maybe create a `stats` attribute flag in `mr-app` and then add this to the scene when it's true?

takahirox:
> Maybe create a stats attribute flag in mr-app and then add this to the scene when it's true?

That sounds reasonable to me.

takahirox:
By the way

```javascript
this.stats.begin();
for (const system of this.systems) {
    system._update(deltaTime, frame);
}
this.stats.end();
```

About [here](https://github.com/Volumetrics-io/mrjs/blob/0ac02418a0073f35c2dca2ea22eb98248de37d5d/src/core/MRApp.js#L471-L479), this wouldn't be better?

```
this.stats.update();
for (const system of this.systems) {
    system._update(deltaTime, frame);
}
```

If I'm right, the current one measures the systems elapsed time while the new one measures the time between two frames. It would be more accurate fps number, won't it?

takahirox:
I started think that showing WebGL related stats in the panel to investigate the performance bottlenecks similar to A-Frame stats panel

![image](https://github.com/Volumetrics-io/mrjs/assets/7637832/9842a09b-4277-459e-925b-20245e8480bc)

PR #474, opened 2024-03-02
Introduce stats attribute to mr-app

takahirox:
From https://github.com/Volumetrics-io/mrjs/issues/441#issuecomment-1958902374 and the following comments.

**Problem**

Stats panel is useful to check the runtime performance and memory consumption. Currently it's enabled if mr-app debug mode. But in debug mode there are many overheads for helping debug and they may be costly. We don't have a way to check the stats in production mode.

**Change for solution**

Separate stats panel from debug mode amd introduce a new stats attribute to mr-app.

Stats panel will be enabled by setting stats attribute like `<mr-app stats="true">`.

This change allows to enable stats in production mode.

**Breaking change**

Stats panel will no longer be enabled just with debug mode like `<mr-app debug="true">`. You need to explicitly add stats attribute if you want to enable it in debug mode like `<mr-app debug="true" stats="true">`.

**Note**

Stats support in immersive mode will be in another PR, not in this PR.

### DIFF
--- a/src/core/MRApp.js
+++ b/src/core/MRApp.js
@@ -214,11 +214,15 @@ export class MRApp extends MRElement {
             }
         }
 
-        if (this.debug) {
+        const statsEnabled = this.getAttribute('stats');
+
+        if (statsEnabled) {
             this.stats = new Stats();
             this.stats.showPanel(0); // 0: fps, 1: ms, 2: mb, 3+: custom
             document.body.appendChild(this.stats.dom);
+        }
 
+        if (this.debug) {
             const orbitControls = new OrbitControls(this.camera, this.renderer.domElement);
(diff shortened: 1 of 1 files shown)


michaelthatsit:
@takahirox is this PR ready? didn't see a review request but don't want it to go ignored.

takahirox:
Yes, ready. I just didn't know that I must assign reviewers when ready.

michaelthatsit:
> Yes, ready. I just didn't know that I must assign reviewers when ready.

Ah no worries. in general just request me or Hannah. I'm going through each open one now.

michaelthatsit (approved):
I noticed your other PR addresses the fact that the stats panel doesn't update in this PR. in the future feel free to combine such updates.

PR #481, opened 2024-03-06
trigger systems for entities instead of manual flags

hanbollar:
core code setup is done, heading back to the media pr to test in headset


hanbollar:
pausing on this while working through bringing the manual documentation back up to date now that we're formalizing the review process

hanbollar:
havent checked this over yet - @michaelthatsit said he might look this over today tho - a few other things might be pending before i turn off wip so i'll take his pass and then a few of my own before going through full review cycle for merging

michaelthatsit:
@hanbollar ah I think I may have been unclear in the original request. I was suggesting we deprecate `needsSystemUpdate` in favor of the `eventUpdate` function i implemented in MRSystem [here](https://github.com/Volumetrics-io/mrjs/blob/c940ab4d05bfbd728ec4c8969ba624ad67bf881e/src/core/MRSystem.js#L158)

[Here's](https://github.com/Volumetrics-io/mrjs/blob/c940ab4d05bfbd728ec4c8969ba624ad67bf881e/src/core/componentSystems/PhysicsSystem.js#L52) how I'm implementing it in the PhysicsSystem for collision shape updates

michaelthatsit:
So I wouldn't implement new events. Instead move the conditional updates into EventUpdate for each system.

hanbollar:
up and running - charging headset then i'll test there for the enter/exitvr triggers as well but overall seems to be running nicely

hanbollar:
since this is an older pr - manually adding pr checklist here:

## Required to Merge

- [ ] **BREAKING CHANGE**

hanbollar:
> up and running - charging headset then i'll test there for the enter/exitvr triggers as well but overall seems to be running nicely

tested in headset with the different examples - seems all good

@michaelthatsit lmk if you find anything ~

michaelthatsit:
@hanbollar It looks like we're getting a pretty noticeable perf hit in-headset. Are you sure you're not accidentally triggering a per-frame update for geometry and material updates?

hanbollar:
> @hanbollar It looks like we're getting a pretty noticeable perf hit in-headset. Are you sure you're not accidentally triggering a per-frame update for geometry and material updates?

we shouldnt be at all - let me add some consoles to check if anything is triggering them directly and get back to you to confirm, but the update functions themselves being empty should keep that from happening

michaelthatsit:
> we shouldnt be at all - let me add some consoles to check if anything is triggering them directly and get back to you to confirm, but the update functions themselves being empty should keep that from happening

Yeah I'm wondering if an event is being triggered every frame and we just hadn't noticed.

hanbollar:
went through it and i dont think it's an issue of items being run every frame, i think there's some items that are being run too often individually regardless of frame (whereas before they were only per frame) - like theyre being hit because of other updates or too often in general due to interaction

the three in question that might be over triggered (ie should be run once a frame instead of everytime theyre trigger) - [here](https://github.com/Volumetrics-io/mrjs/pull/481/files#diff-01c9b3aebb48b836662bdfbc73400faad3960063a2774426d45e89b0866c14d4R334-R335), for [child-updated](https://github.com/Volumetrics-io/mrjs/pull/481/files#diff-01c9b3aebb48b836662bdfbc73400faad3960063a2774426d45e89b0866c14d4R249-R250), and for [mouse-events](https://github.com/Volumetrics-io/mrjs/pull/481/files#diff-01c9b3aebb48b836662bdfbc73400faad3960063a2774426d45e89b0866c14d4R217-R218)

i think mouse over has to happen as an individual event, but the other two might be better as per-frame updates - or if there's a way we can ignore the in between and only run their updates per frame if theyre triggered for an entity - thoughts?

hanbollar:
im going to keep digging on this

michaelthatsit:
Using the perf profiler in chrome it looks like the `TextSystem.update` is the culprit! looks like some of the logic in `update` might need to be moved into `eventUpdate` as well. Most probably `updateStyle`

hanbollar:
good catch! - i'll add that change to this pr

michaelthatsit:
`MaterialSystem` and `GeometrySystem` don't show up on the profiler at all tho which is great!

hanbollar:
kk @michaelthatsit checked in headset again - /should/ be good now - but lmk if there's anything lingering you'd want me to tackle in this one

michaelthatsit (approved):

PR #494, opened 2024-03-12
Fix clipping targets

takahirox:
**Problem**

Clipping system is designed to apply the clipping only the entities that are not masking system targets as optimization. But currently due to a bug the entities that are masking system targets can be also the targets of the clipping system.

This problem can cause a bad performance impact because clippings are done more than needed.

**Change for solution**

Fix the bug.

In the change for the bug fix, introducing `Object3D.userData.isEntityObject3DRoot` to distinguish whether an Three.js `Object3D` is a root `Object3D` of MR entity to handle clipping targets properly.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/MREntity.js
+++ b/src/core/MREntity.js
@@ -52,6 +52,7 @@ export class MREntity extends MRElement {
         });
 
         this.object3D = new THREE.Group();
+        this.object3D.userData.isEntityObject3DRoot = true;
         this.object3D.userData.bbox = new THREE.Box3();
         this.object3D.userData.size = new THREE.Vector3();
 
--- a/src/core/componentSystems/ClippingSystem.js
+++ b/src/core/componentSystems/ClippingSystem.js
@@ -99,15 +99,28 @@ export class ClippingSystem extends MRSystem {
      * @function
      * @description Helper method for `onNewEntity`. Actually applies the clipping planes to the material setup for rendering.
      * Uses threejs in the background following https://threejs.org/docs/?q=material#api/en/materials/Material.clippingPlanes
-     * @param {object} object - the object3D item to be clipped
+     * @param {MREntity} entity - the entity to be clipped
      * @param {MRClippingGeometry} clipping - the clipping information to be passed to the material
      */
-    applyClipping(object, clipping) {
-        if (!object.isMesh) {
+    applyClipping(entity, clipping) {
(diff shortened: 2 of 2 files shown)


src/core/MREntity.js
```diff
@@ -52,6 +52,7 @@ export class MREntity extends MRElement {
         });
 
         this.object3D = new THREE.Group();
+        this.object3D.userData.isEntityObject3DRoot = true;
```
takahirox, line 55: I introduced `Object3D.userData.isEntityObject3DRoot` to allow to check whether an object is the root object of an entity.

I needed this to traverse the object3Ds only associated with an entity, like

```javascript
class traverse = (obj) => {
  something(obj).
  for (const child of obj.children) {
    if (!obj.userData.isEntityObject3DRoot) {
      traverse(child);
    }
  }
};
traverse(entity.object3D);
```

Honestly I think there might be a more elegant approach. Please suggest better ideas if you come up with.

hanbollar: one shorter solution is leaning into the assumption that if object3D is a group, then object3D.children[0] is the root in question

reasoning - when the entities are set up it'll be either the first background object for any mrdiv or the 'object' that should be representing the entity itself given how we instantiate the items in the 'connected' functions - meaning it would actually be the root object

but i'm hesitant to lean towards that as it becomes more overhead long term - and if a user does any of their own javascript modifications or we change the architecture in future, that logic will become a huge edge case

i think what you have, though it's a flag, is straight forward enough for now

in future if/when we can lean into traversing over the entities directly instead of the object3Ds (if that is ever possible) then that reduces our need for the flag, but that's not worth a dive through at the moment.

hanbollar: if you're meaning the flag to represent the direct object3d of the entity root, then doing a traversal over the entities instead of the object3d items would be the only alternative, but my one concern would be from the object3ds being overly nested based on the entity hierarchy (you end up with factorially redundant traversals)

so same thing - if it ends up being cleaner great, but the flag approach is straightforward for now

michaelthatsit:
I'm looking over the code now but just tested in headset, perf seems great and clipping is working as expected.

michaelthatsit (approved):
This is a good solution. we could probably polish it in the future.

PR #499, opened 2024-03-13
Optimize Clipping system

### LINKED ISSUE #466: Reduce the number of clipping planes
**Problem**

Clipping systems seems to be slow. Perhaps one of the reasons would be clipping planes are created more than they need to be.

12 planes seems to be created for a panel, looks like (unexpectedly?) duplicated. Testing 12 times with planes per vertex would be costly.

image

**Solution**

I think 6 planes (6 faces in a box), or less if possible, would be good enough.

takahirox:
## Linking

Fixes #466

## Problem

Currently clipping system is a bit more inefficient because it assigns 12 panels to an entity although 6 panels may be good enough and it unnecessarily creates a new `BufferGeometry` in `update()`.

## Changes for solution

Optimize and clean up.

- Assigns 6 panels to an entity
- Avoid new `BufferGeometry` creation in `update()`
- Remove some duplicated codes

## Breaking changes

Limiting `clipping.geometry` to one segment `BoxGeometry` for simplicity.

@michaelthatsit 

The current implementation seems that support for complex ClippingGeometry shapes is also being considered. However, there are also some codes that seem to assume BoxGeometry. In this PR, I would like to propose changing the specification to assume BoxGeometry for simplicity for now. Introducing complex shape and logic makes it easier for bugs to creep in, such as in the case of 12 planes. We can consider how to support for complex shapes in the future. I believe that keeping things simple will also make it easier to make improvements in the future.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/componentSystems/ClippingSystem.js
+++ b/src/core/componentSystems/ClippingSystem.js
@@ -6,6 +6,8 @@ import { MREntity } from 'mrjs/core/MREntity';
 import { MRClippingGeometry } from 'mrjs/dataTypes/MRClippingGeometry';
 import { MRVolume } from '../entities/MRVolume';
 
+const PLANE_NUM = 6;
+
 /**
  * @class ClippingSystem
  * @classdesc   This system supports 3D clipping following threejs's clipping planes setup.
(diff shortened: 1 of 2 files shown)


hanbollar (approved):
code-wise everything looks good - leaving to @michaelthatsit for final review ~

michaelthatsit:
Clipping behaves as expected! and simplifying to BoxGeometry makes sense, supporting others geometries would be valuable but can remain out of scope for now.

michaelthatsit (approved):

PR #507, opened 2024-03-16
Fix Masking System

### LINKED ISSUE #505: Clipping and/or masking not applied to some elements
Just caught this issue on the landing page. as you can see, the blue example tiles aren't being masked or clipped. 

https://github.com/Volumetrics-io/mrjs/assets/19628643/1e56401a-8b00-448a-87fd-25f1b62c62ad

### LINKED ISSUE #509: Masking system race condition
From @michaelthatsit on Discord

image

He found that sometimes(?) text is not masked. Perhaps race condition?

takahirox:
## Linking

Fixes #505
Fixes #509

## Problems

Problem 1:

Perhaps due to the recent changes to mapping and clipping systems, background of mr div entities seem not to be masked.

The root issue seems to be in masking system. Child objects associated with an entity need to be masked but we have been missing to set up their materials for stencil.

Problem 2:

There seems to be a race condition. Sometimes children of panels are not masked.

## Change for fix

Fix for the problem 1:

Traverse the objects associated with an entity and set up their materials.

This commit introduces a new `traverseObjects` method to `MREntity` that runs the passed through function on the objects associated with an entity as a helper method because such traversal is needed in masking and clipping systems, and it would be used from more many places in the future.

Fix for the problem 2:

Take into account that a child entity of panel can be added after parent panels are registered in masking system. Check whether there are panels in parents and set up material for stencil if there are, similar to what clipping system does.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/MREntity.js
+++ b/src/core/MREntity.js
@@ -420,6 +420,23 @@ export class MREntity extends MRElement {
             child.traverse(callBack);
         }
     }
+
+    /**
+     * @function
+     * @description Runs the passed through function on the objects associated with this Entity
+     * @param {Function} callBack - the function to run recursively.
+     */
+    traverseObjects(callBack) {
+        const traverse = (object) => {
+            callBack(object);
+            for (const child of object.children) {
+                if (!child.userData.isEntityObject3DRoot) {
+                    traverse(child);
+                }
+            }
+        };
+        traverse(this.object3D);
+    }
 }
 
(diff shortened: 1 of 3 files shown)


takahirox:
Please don't merge yet because I will add the fix of #509 in this PR. I will let you know when ready.

michaelthatsit (approved):
Works and perf is great!

hanbollar (approved):
lgtm ~

Issue #493, opened 2024-03-11
LayoutSystem seems to be slow


takahirox:
**Problem**

LayoutSystem seems to be slow. Especially, [getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) seems to be slow.

Because of it, even just with a panel and empty 32 mr-div entities fps number drops from 90 fps in immersive mode on my Quest 3.

![image](https://github.com/Volumetrics-io/mrjs/assets/7637832/25d2dc3c-0c01-47e6-972f-277d6c4a3a4d)

**Potential solution ideas**

I'm not familiar with LayoutSystem yet but we might do

1. Update the system only when needed if possible (currently it's called every frame?)
2. Call getBoundingClientRect() only when needed if possible
3. Use another API from getBoundingClientRect() if possible ((Layout related) DOM API is costly?)

**Additional information**

https://github.com/TanStack/virtual/issues/359 may help

takahirox:
As I repeated the performance tests, I started to feel that the slowness of the LayoutSystem could not be ignored. `getBoundingClientRect()` seems to be the root cause so I'm thinking of detecting the change, for example

https://stackoverflow.com/questions/40251082/an-event-or-observer-for-changes-to-getboundingclientrect

but am not sure yet if it always works.

takahirox:
If I'm right `Element.getBoundingClientRect()` (or reflow caused by it) can drop the fps number significantly. Not sure if it's Quest 3 or Meta browser specific. It may be good to test on other platforms, and report to Meta if we found that it is their platform specific problem.

PR #508, opened 2024-03-16
MRStats: Stats panel in immersive mode

### LINKED ISSUE #441: `mr-stats` element for tracking perf in XR

takahirox:
This PR adds `MRStats(mr-stats)` to allow the stats panel even in immersive mode.

This PR is still draft, but let me share for some questions (see the inline comments) and also for people who already want to use in local test.

Screenshot:

dec8cc33c34f24a65f0bc684a98f4e16

--

The PR comment will be written when the PR is ready.

## Linking

Fixes #441

## Problem

*Description of the problem including potential code and/or screenshots as an example*

## Solution

*Quick explanation of change to be done*

### Breaking Change

*If this is a breaking change describe the before and after and why the change was necessary*

## Notes

Out of scope in this PR:
* Enhance the stats panel info (eg: add WebGL stats). That would be in another future PR.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/MRApp.js
+++ b/src/core/MRApp.js
@@ -25,6 +25,7 @@ import { SkyBoxSystem } from 'mrjs/core/componentSystems/SkyBoxSystem';
 import { TextSystem } from 'mrjs/core/componentSystems/TextSystem';
 import { AudioSystem } from 'mrjs/core/componentSystems/AudioSystem';
 import { PanelSystem } from 'mrjs/core/componentSystems/PanelSystem';
+import { StatsSystem } from 'mrjs/core/componentSystems/StatsSystem';
 import MRUser from 'mrjs/core/user/MRUser';
 
 ('use strict');
@@ -113,7 +114,15 @@ export class MRApp extends MRElement {
         this.observer = new MutationObserver(this.mutationCallback);
         this.observer.observe(this, { attributes: true, childList: true });
 
+        if (this.getAttribute('stats')) {
+            const statsEl = document.createElement('mr-stats');
+            statsEl.setAttribute('data-comp-anchor', 'type: plane; label: wall;');
+            statsEl.setAttribute('data-position', '0 0.2 0.01');
+            this.appendChild(statsEl);
+        }
+
         // order matters for all the below system creation items
+        this.statsSystem = new StatsSystem();
         this.panelSystem = new PanelSystem();
         this.layoutSystem = new LayoutSystem();
         this.textSystem = new TextSystem();
@@ -213,14 +222,6 @@ export class MRApp extends MRElement {
             for (const layer of this.layers) {
                 this.camera.layers.enable(layer);
             }
-        }
-
-        const statsEnabled = this.getAttribute('stats');
-
-        if (statsEnabled) {
-            this.stats = new Stats();
-            this.stats.showPanel(0); // 0: fps, 1: ms, 2: mb, 3+: custom
-            document.body.appendChild(this.stats.dom);
         }
 
         if (this.debug) {
@@ -446,12 +447,6 @@ export class MRApp extends MRElement {
         // ----- grab important vars ----- //
 
         const deltaTime = this.clock.getDelta();
-
-        // ----- Update stats if enabled ----- //
-
-        if (this.stats) {
-            this.stats.update();
-        }
 
         // ----- Update needed items ----- //
 
--- a/src/core/componentSystems/StatsSystem.js
+++ b/src/core/componentSystems/StatsSystem.js
@@ -1,0 +1,59 @@
+import { mrjsUtils } from 'mrjs';
+import { MRSystem } from 'mrjs/core/MRSystem';
+import { MRStats } from 'mrjs/core/entities/MRStats';
+
+import Stats from 'stats.js';
+
+import { HTMLMesh } from 'three/addons/interactive/HTMLMesh.js';
+import { InteractiveGroup } from 'three/addons/interactive/InteractiveGroup.js';
+
+// TODO: JSDoc
+
+export class StatsSystem extends MRSystem {
+    constructor() {
+        super(false);
+        this.statsEntities = [];
+    }
+
+    update() {
+        for (const entity of this.statsEntities) {
+            if (entity.stats === null && this.app) {
+                entity.stats = new Stats();
+                entity.stats.showPanel(0); // 0: fps, 1: ms, 2: mb, 3+: custom
+                document.body.appendChild(entity.stats.dom);
+
+                // Note: InteractiveGroup API is refactored in the newer Three.js rev.
+                //       Don't forget to update the followings when upgrading Three.js
+                entity.interactiveGroup = new InteractiveGroup(this.app.renderer, this.app.camera);
+                entity.object3D.add(entity.interactiveGroup);
+
+                // Note: HTMLMesh seems to have a bug that image is not displayed at a correct position.
+                //       in Three.js r161 or older.
+                // TODO: Upgrade Three.js to fix
+
+                // Question: Circle pointer doesn't seem to be displayed on the Stats panel object
+                //           and the clickable area seems to be a difference place from the panel object, why?
+                // TODO: Fix it
+                entity.statsMesh = new HTMLMesh(entity.stats.dom);
+                entity.interactiveGroup.add(entity.statsMesh);
+            }
+
+            if (entity.stats) {
+                entity.stats.update();
+                if (mrjsUtils.xr.isPresenting) {
+                    entity.statsMesh.visible = true;
+                    entity.statsMesh.material.map.update();
+                } else {
+                    entity.statsMesh.visible = false;
+                }
+            }
+        }
+    }
+
+    onNewEntity(entity) {
+        // Question: How can we detect the removal of entities?
+        if (entity instanceof MRStats) {
+            this.statsEntities.push(entity);
+        }
+    }
+}
--- a/src/core/entities/MRStats.js
+++ b/src/core/entities/MRStats.js
@@ -1,0 +1,20 @@
+import { MRDivEntity } from '../MRDivEntity';
+
+// TODO: JSDoc
+
+export class MRStats extends MRDivEntity {
+    constructor() {
+        super();
+        this.ignoreStencil = true;
+        this.object3D.name = 'stats';
+
+        // Three.js InteractiveGroup has a dependency with WebGLRenderer and Camera
+        // instances but they can't be accessed from here so these objects are initialized
+        // in StatsSystem instead.
+        this.stats = null;
+        this.interactiveGroup = null;
+        this.statsMesh = null;
+    }
+}
+
+customElements.get('mr-stats') || customElements.define('mr-stats', MRStats);
--- a/src/index.js
+++ b/src/index.js
@@ -42,6 +42,7 @@ import './core/entities/MRLight';
 import './core/entities/MRModel';
 import './core/entities/MRPanel';
 import './core/entities/MRSkyBox';
+import './core/entities/MRStats';
 import './core/entities/MRTextArea';
 import './core/entities/MRTextField';
 // CORE - COMPONENT-SYSTEMS
@@ -56,6 +57,7 @@ import './core/componentSystems/MaskingSystem';
 import './core/componentSystems/MaterialStyleSystem';
 import './core/componentSystems/PhysicsSystem';
 import './core/componentSystems/SkyBoxSystem';
+import './core/componentSystems/StatsSystem';
 import './core/componentSystems/TextSystem';
 
 // EXPORTS


takahirox:
I'm not so familiar with our ECS design yet. I'm happy if you give any comments for making the implementation in this PR aligned with it @michaelthatsit @hanbollar

src/core/componentSystems/StatsSystem.js
```diff
@@ -0,0 +1,59 @@
...
+                //       Don't forget to update the followings when upgrading Three.js
+                entity.interactiveGroup = new InteractiveGroup(this.app.renderer, this.app.camera);
+                entity.object3D.add(entity.interactiveGroup);
+
+                // Note: HTMLMesh seems to have a bug that image is not displayed at a correct position.
+                //       in Three.js r161 or older.
+                // TODO: Upgrade Three.js to fix
+
+                // Question: Circle pointer doesn't seem to be displayed on the Stats panel object
+                //           and the clickable area seems to be a difference place from the panel object, why?
```
takahirox, line 35: @michaelthatsit @hanbollar Do you have any clue for this problem?

src/core/componentSystems/StatsSystem.js
```diff
@@ -0,0 +1,59 @@
...
+                document.body.appendChild(entity.stats.dom);
+
+                // Note: InteractiveGroup API is refactored in the newer Three.js rev.
+                //       Don't forget to update the followings when upgrading Three.js
+                entity.interactiveGroup = new InteractiveGroup(this.app.renderer, this.app.camera);
+                entity.object3D.add(entity.interactiveGroup);
+
+                // Note: HTMLMesh seems to have a bug that image is not displayed at a correct position.
+                //       in Three.js r161 or older.
+                // TODO: Upgrade Three.js to fix
```
takahirox, line 32: @michaelthatsit @hanbollar What do you think of upgrading Three.js to the latest?

hanbollar:
> I'm not so familiar with our ECS design yet. I'm happy if you give any comments for making the implementation in this PR aligned with it @michaelthatsit @hanbollar

I'll add a few comments in the code with some explanations to help with this a little bit - the main bullets are:
- MRStats should be MREntity not MRDivEntity
- use the entity's 'connected' function to create itself instead of the update

src/core/entities/MRStats.js
```diff
@@ -0,0 +1,20 @@
+import { MRDivEntity } from '../MRDivEntity';
+
+// TODO: JSDoc
+
+export class MRStats extends MRDivEntity {
```
hanbollar, line 5: MRStats should be an `MREntity` instead of `MRDivEntity`

think of `MRDivEntity` items to be (except for MRModel) the UI elements - that is theyre the ones that are situated on a panel almost as if that panel is a website in 3D space

Since we want MRStats to be its own object regardless of a panel and to act for the app as a whole - `MREntity` is better

src/core/MRApp.js
```diff
@@ -113,7 +114,15 @@ export class MRApp extends MRElement {
         this.observer = new MutationObserver(this.mutationCallback);
         this.observer.observe(this, { attributes: true, childList: true });
 
+        if (this.getAttribute('stats')) {
+            const statsEl = document.createElement('mr-stats');
+            statsEl.setAttribute('data-comp-anchor', 'type: plane; label: wall;');
+            statsEl.setAttribute('data-position', '0 0.2 0.01');
+            this.appendChild(statsEl);
+        }
+
```
hanbollar, line 123: this would be better off in the constructor of `StatsSystem` as we should only have a stats system if `this.getAttribute('stats')` is true

src/core/componentSystems/StatsSystem.js
```diff
@@ -0,0 +1,59 @@
...
+
+export class StatsSystem extends MRSystem {
+    constructor() {
+        super(false);
+        this.statsEntities = [];
+    }
+
+    update() {
+        for (const entity of this.statsEntities) {
+            if (entity.stats === null && this.app) {
```
hanbollar, line 20: this section should be in the entity's `connected` function - this is the function that gets called when the entity is 'connected to an app (meaning the app exists and <mr-stats> is in the html within it)

[mr-hyperlink's use of connected ](https://github.com/Volumetrics-io/mrjs/blob/main/src/core/entities/MRHyperlink.js#L39)is a good example of this

hanbollar: to clarify, when i say 'this section' here i mean within the `if (entity.stats...) { }`

src/core/componentSystems/StatsSystem.js
```diff
@@ -0,0 +1,59 @@
...
+                }
+            }
+        }
+    }
+
+    onNewEntity(entity) {
+        // Question: How can we detect the removal of entities?
+        if (entity instanceof MRStats) {
+            this.statsEntities.push(entity);
+        }
```
hanbollar, line 57: at the moment, we do not remove entities from this set once theyre added

in future if we ever have live entity deletion that will need to happen, but it will probably be by use of an additional system we add to handle all the locations an entity is included

----

an additional note:

every MRSystem has a 'this.registry' which is a set for every entity associated with a system

`this.statsEntities.push(entity)` -->  `this.registry.add(entity);`

src/core/componentSystems/StatsSystem.js
hanbollar: not entirely sure at first glance - i'll take a quick look in headset and get back to you on that

hanbollar: my first thought would be - it might be due to the following not interacting with each other nicely: [document item creation](https://github.com/Volumetrics-io/mrjs/pull/508/files#diff-9a9b14ca4cde264f45d74bc2e9e2926a929c81d6fccf4161133d5fc99d48d3faR117-R122) and [document child append](https://github.com/Volumetrics-io/mrjs/pull/508/files#diff-2dec46b40149b872e74789e7c24f5881c4b2443f3246f06414bbf697819e55b2R23)

where you have both the html item and the threejs version of it showing up (with only one being recognized as the clickable item - and with the adjustment for position from [here](https://github.com/Volumetrics-io/mrjs/pull/508/files#diff-9a9b14ca4cde264f45d74bc2e9e2926a929c81d6fccf4161133d5fc99d48d3faR120) )

----

but i'm not 100% sure it makes full sense that that will cause the issue you're experiencing with the rest of what you currently have implemented

hanbollar:
regarding your question about [`mr-stats` appearing/interacting weirdly](https://github.com/Volumetrics-io/mrjs/pull/508#discussion_r1527171123) - make the changes i mentioned in the comments above for ECS and if the problem still persists i'll take a look again 👍 

hope that helps a bit for now ~

src/core/componentSystems/StatsSystem.js
hanbollar: the other thing is - since this pr is not entirely following the ECS setup yet in the system, it might be hitting a weird case where on creation, some of the entity/dom items arent getting interpreted in the overall setup properly

once the `mr-stats` and `MRSystem` are in a better ECS setup, it'll probably be clearer as to why this issue is happening

src/core/componentSystems/StatsSystem.js
takahirox: Any proper way to access `app` from `connected()`? Because of dependency with `app` I wrote this initialization in system.

src/core/MRApp.js
takahirox: Not a final conclusion yet but I'm thinking of also allowing users to write `<mr-stats>` explicitly in their contents, like the following, so initializing it outside of the system constructor for now.

```
<mr-stats data-comp-anchor="..." " data-position="..."/>
```

hanbollar: i think that's a good option since it aligns much more with the idea of how we're using it as an entity

we can also allow for both - kind of like how MRSkybox you can set it as `<mr-skybox...>` or just directly through the css of the `mr-app`'s background property

src/core/componentSystems/StatsSystem.js
hanbollar: not directly, but a decent workaround is: 

you leave the main 'constructor' the implementation that doesnt rely on the dependency directly into the entity's own 'constructor' or 'connected' functions (whichever makes most sense in your opinion)

and then you move the part of this 'constructor' implementation that relies on mr-app to the `onNewEntity` section (kind of like [how we do it in LayoutSystem](https://github.com/Volumetrics-io/mrjs/blob/main/src/core/componentSystems/LayoutSystem.js#L34))

src/core/MRApp.js
hanbollar: also - something that might be helpful as your implementation will potentially be creating <mr-stats> on the fly as well
- see the note i had mentioned about MRHyperlink (since it makes its dom item directly in the entity)
- and then i forgot to link this for MRSkybox: https://github.com/Volumetrics-io/mrjs/blob/main/src/core/MRApp.js#L248 <-- this is a good reference for adding in elements on the fly into the MRApp s.t. it works nicely with our pre-existing system

src/core/componentSystems/StatsSystem.js
takahirox: And, with the newer Three.js rev, the initialization will have a dependency with `app.user` initialized at https://github.com/Volumetrics-io/mrjs/blob/15cc832a595506488906a4d97fe9c2055ea1e86d/src/core/MRApp.js#L125 Any good practice to initialize a resource with `app.user`?

hanbollar: that's a good question - for reference:
- systems are only there if an app exists: [ref](https://github.com/Volumetrics-io/mrjs/blob/main/src/core/MRApp.js#L108)
- [engine-started](https://github.com/Volumetrics-io/mrjs/blob/979ea75652b135e175296fb05b4159b42a9d7a96/src/utils/Physics.js#L24) is triggered by the physics world being created properly by rapier
- which triggers the creation of the [remaining systems](https://github.com/Volumetrics-io/mrjs/blob/main/src/core/MRApp.js#L124)
- each system then runs through the scene for each entity and touches them one by one
that is - since `<mr-app>` exists, we hit the `onNewEntity` function within whatever system is being called
- meaning it is guaranteed the `app` is ready when the `onNewEntity` function for the system is hit

let me know if that makes sense - we still need to add better documentation for how our ECS rundown works in MRjs in docs.mrjs.io, but I tried to use code-callouts for reference

hanbollar: [for this question](https://github.com/Volumetrics-io/mrjs/pull/508#discussion_r1529447438) - this just means that it's a better case to have that part of the entity initialized through `onNewEntity` of the associated system

there are three places where an entity can/should be setup:
- its own constructor
- its own `connected` function
- a system's `onNewEntity` function

the more scope that is needed, the more you go down the bulleted list

takahirox:
Note that the stats panel itself in immersive mode may have a performance impact because canvas for the stats panel needs to uploaded to texture every frame. Actually just displaying the stats panel in immersive mode drops frame counter a few fps counter number on my Oculus Quest 3. It's still useful, but be careful if you want to test very sensitive performance stuffs (eg: check whether FPS counter is stable at 90).

takahirox:
If we want more performant stats panel, we may refer to the one in [WebXR samples](https://immersive-web.github.io/webxr-samples/). It uploads only the needed sub data, the visual quality would be much simpler tho.

takahirox:
I found that Three.js `HTMLMesh` used in this PR uses `Element.getBoundingClientRect()`. It may be another cause that drops the FPS number. Refer to #493

michaelthatsit:
> If we want more performant stats panel, we may refer to the one in [WebXR samples](https://immersive-web.github.io/webxr-samples/). It uploads only the needed sub data, the visual quality would be much simpler tho.

The immersive web one would be fine, obviously I'd love to have something a little more aesthetically pleasing and interactive. 

Would it be better for us to implement our own stats panel using MRjs? We could extend MRTextEntity and add some simple update logic, obviously there would be no graph, just numbers.

takahirox:
> Would it be better for us to implement our own stats panel using MRjs? We could extend MRTextEntity and add some simple update logic, obviously there would be no graph, just numbers.

That would be good. Actually I started to think we need to revisit the implementation of this PR.

The reason why I use Three.js HTMLMesh and InteractiveGroup is I thought it may be easy to implement. But I noticed that it may have a significant performance impact (due to the use of Element.getBoundingClientRect() in HTMLMesh and/or the need of entire texture data upload every frame), for example on skybox example it seems to run nearly at 90 fps without this stats panel but the fps number drops to about 45 with this immersive stats panel. Stats panel that can have a big performance impact doesn't really make sense. We need more performant one.

takahirox:
Closing because slow stats panel doesn't really make sense. I would like to try simpler one with MRText in another PR.

Issue #506, opened 2024-03-16
Allow the ability to save the canvas as a PNG with transparency


lobau:
Sometimes it's nice to be able to right click the canvas, select "Save image as" and get a PNG of the MRjs scene, with preserved transparency.

By default, three.js doesn't preserve the drawing buffer (for performance reasons), so you have to enable it ( https://stackoverflow.com/questions/15558418/how-do-you-save-an-image-from-a-three-js-canvas )

```js
new THREE.WebGLRenderer({
    preserveDrawingBuffer: true 
});
```

It would be nice to pass it through `<mr-app>`, maybe as a boolean:

```html
<mr-app preserve-buffer="true">…</mr-app>
```

takahirox:
Adding opt-in sounds good to me

hanbollar:
following up from the convo here: https://github.com/Volumetrics-io/mrjs/pull/511#discussion_r1529381795

@lobau for the toggling you're requesting - is there a specific browser you were hitting your issue? i've been finding i have it by default in my setup and want to cross check why it's different than your experience

<img width="372" alt="Screenshot 2024-03-18 at 4 31 53 PM" src="https://github.com/Volumetrics-io/mrjs/assets/19377312/d6913045-1c6d-45bf-9653-a69acf618bb9">

is it specifically about the alpha preservation?

lobau:
@hanbollar currently if you save the image, it will be a blank PNG. The action is visible in the browser (because it's a canvas element), but the canvas element is empty.

hanbollar:
discussed - looks like chrome and any chrome-based browsers (aka most newer ones including duckduckgo) have something inline in the chrome state s.t. save-as just works

firefox doesnt have it - using that as a good 'debugger' for this to confirm the feature with the flag works

PR #511, opened 2024-03-18
add toggle in `mr-app` for `preserve-drawing-buffer`

hanbollar:
## Linking

Fixes https://github.com/Volumetrics-io/mrjs/issues/506

## Problem

Sometimes it's nice to be able to right click the canvas, select "Save image as" and get a PNG of the MRjs scene, with preserved transparency.

## Solution

add in an ability to toggle the `preserve-drawing-buffer` flag for the threejs renderer

### Breaking Change

adds the `<mr-app>` flag `preserve-drawing-buffer`

## Notes

            // There's issues in the timing to enable taking screenshots of threejs scenes unless you have direct access to the code.
            // Using the preserveDrawingBuffer to ignore timing issues is the best approach instead. Though this has a performance hit,
            // we're allowing it to be enabled by users when necessary.
            //
            // References:
            // https://stackoverflow.com/questions/15558418/how-do-you-save-an-image-from-a-three-js-canvas
            // https://stackoverflow.com/questions/30628064/how-to-toggle-preservedrawingbuffer-in-three-js

------------

## Required to Merge

~~- [ ] **TEST IN HEADSET** - [main dev-testing-example](https://github.com/Volumetrics-io/mrjs/tree/main/samples/index.html) and any of the other [examples](https://github.com/Volumetrics-io/mrjs/tree/main/samples/examples) still work as expected~~
~~- [ ] **VIDEO** - if this pr changes something visually - post a video here of it in headset-MR and/or on desktop (depending on what it affects) for the reviewer to reference.~~
- [x] **BREAKING CHANGE**
  - link the pr of the documentation repo here: *https://github.com/Volumetrics-io/documentation/pull/26*


src/core/MRApp.js
```diff
@@ -202,6 +202,16 @@ export class MRApp extends MRElement {
...
         this.renderer.localClippingEnabled = true;
+        if (this.getAttribute('preserve-drawing-buffer') ?? false) {
+            // There's issues in the timing to enable taking screenshots of threejs scenes unless you have direct access to the code.
+            // Using the preserveDrawingBuffer to ignore timing issues is the best approach instead. Though this has a performance hit,
+            // we're allowing it to be enabled by users when necessary.
+            //
+            // References:
+            // https://stackoverflow.com/questions/15558418/how-do-you-save-an-image-from-a-three-js-canvas
+            // https://stackoverflow.com/questions/30628064/how-to-toggle-preservedrawingbuffer-in-three-js
+            this.renderer.preserveDrawingBuffer = true;
```
takahirox, line 213: `preserveDrawingBuffer` flag must be passed via the `WebGLRenderer` constructor.

https://threejs.org/docs/#api/en/renderers/WebGLRenderer

hanbollar: good catch with that

---

very odd, i feel like im not seeing any difference using it in constructor, outside, or even not having it at all

i wonder if it's a browser thing - i'll look into this and follow back up for why the discrepancy as we might not even need the flag then if it works?

weird...

hanbollar: discussed - following up per [conversation](https://github.com/Volumetrics-io/mrjs/issues/506#issuecomment-2007921050) for why issue wasnt showing up on my end - and [added note about chromium base](https://github.com/Volumetrics-io/documentation/pull/26/commits/9f623225d10f8ad7e0d99e0e815032ab8bca2dde) to the documentation pr

hanbollar: going to continue to implement a working fix ~

hanbollar:
recap of fix since the last review
- moved s.t. `preserve-drawing-buffer` is properly in the constructor --> had to move renderer creation to do so, but works fine with that setup
- added a debug.html bc it got annoying adding and removing the flags over and over --> this is a small thing for now, in future we'll probably make a more robust html for it
- added the html to testing
- updated the documentation pr to reflect the browser note as well (regarding need or no-need for the flag itself)

@takahirox @michaelthatsit lmk if there's anything else that might be blocking for this pr being merged ~

src/core/MRApp.js
```diff
@@ -193,6 +195,19 @@ export class MRApp extends MRElement {
...
+            antialias: true,
+            alpha: true, 
+            // There's issues in the timing to enable taking screenshots of threejs scenes unless you have direct access to the code.
+            // Using the preserveDrawingBuffer to ignore timing issues is the best approach instead. Though this has a performance hit,
+            // we're allowing it to be enabled by users when necessary.
+            //
+            // References:
+            // https://stackoverflow.com/questions/15558418/how-do-you-save-an-image-from-a-three-js-canvas
+            // https://stackoverflow.com/questions/30628064/how-to-toggle-preservedrawingbuffer-in-three-js
+            preserveDrawingBuffer: this.getAttribute('preserve-drawing-buffer') ?? false,
```
michaelthatsit, line 209: Maybe we shorten it to `preserve-buffer`? Could just be my aversion to long attribute names.

hanbollar: I agree with the reasoning; however, I'd rather not given this is us just exposing a threejs flag underneath so it's better practice for maintainability to keep them the same - if it was one we made ourselves though 100% shorter is better

michaelthatsit: Fair point. maybe in the future as we expose more renderer options at the attribute level, we can fall back to an all encompassing `data-renderer` flag. something like.

`<mr-app data-renderer="antialias: true; alpha: true; preserveDrawingBuffer: true; ...">

rather than having one for each possible value.

hanbollar: i like that idea - i'll open an issue for tracking and it'll probably stay on the backlog for a bit, but there's definitely other items in there we'd want to expose as well ~

hanbollar: issue ref: #519

PR #515, opened 2024-03-21
Drop MRTextEntity traverse method override

takahirox:
## Linking

Fixes: https://github.com/Volumetrics-io/mrjs/pull/306/files#r1533265296

## Problem

MRTextEntity overrides .traverse() method to traverse .object3D.children (Three.js Object3D collections). But it's a error prone because parent .traverse() traverses .children (HTMLElement collection). Callback function doesn't expect Object3D argument if MRTextEntity.traverse() is called from parent .traverse().

## Solution

Simply drop MRTextEntity.traverse(). Use textEntity.object3D.traverse() instead if object3Ds traverse is needed.

------------

## Required to Merge

- [x] **PASS** - all necessary actions must pass (excluding the auto-skipped ones)
- [ ] **TEST IN HEADSET** - [main dev-testing-example](https://github.com/Volumetrics-io/mrjs/tree/main/samples/index.html) and any of the other [examples](https://github.com/Volumetrics-io/mrjs/tree/main/samples/examples) still work as expected
- [ ] **VIDEO** - if this pr changes something visually - post a video here of it in headset-MR and/or on desktop (depending on what it affects) for the reviewer to reference.
- [x] **TITLE** - make sure the pr's title is updated appropriately as it will be used to name the commit on merge
- [ ] **BREAKING CHANGE**
  - make a pr in the [documentation repo](https://github.com/Volumetrics-io/documentation) that updates the manual docs to match the breaking change
  - note: the docs underneath the `Javascript API` section come automated from the jsdocs inline with the code itself
  - link the pr of the documentation repo here: *#pr*
  - that documentation repo pr must be approved by `@lobau`

PR #516, opened 2024-03-21
Introduce BoundaryVisibilitySystem

### LINKED ISSUE #468: Make objects invisible that are obviously outside of the plane
**Problem**

We want some objects to be shown only when they are on the associated panels in immersive mode. Otherwise we want them to be hidden. (If I understand correctly) we use masking and clipping systems to control them. But unfortunately those systems seems to be costly.

**Suggestion**

We add a new system that checks the positions of objects and associated panels and sets `.visible` to `false` of the entirely objects that are obviously outside of the panels. Perhaps this additional cost in CPU would be much cheaper compared to clipping or masking in GPU.

takahirox:
## Linking

Fixes #468
Requires #515

## Problem

Currently entities outside of their parent panels are masked with stencil or clipped with clipping. But stencil and clipping are costly.

## Change for solution

Introduce a new system named `BoundaryVisibilitySystem` that makes the entities that outside of their parent panels invisible with `.object3D.visible = false`. It should be much cheaper than stencil or clipping.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/componentSystems/BoundaryVisibilitySystem.js
+++ b/src/core/componentSystems/BoundaryVisibilitySystem.js
@@ -1,0 +1,81 @@
+import { MRSystem } from 'mrjs/core/MRSystem';
+import { MRDivEntity } from 'mrjs/core/MRDivEntity';
+import { MRPanel } from 'mrjs/core/entities/MRPanel';
+
+/**
+ * @function
+ * @description Observe a target MRDivEntity and make the associated object visible only if it is in visible position in a root MRDivEntity
+ * @param {MRDivEntity} root
+ * @param {MRDivEntity} target
+ */
+const observe = (root, target) => {
+    // TODO: Callback is fired asynchronously so no guaranteed to be called immediately when the
+    //       visibility from the layout position changes. Therefore, the visibility of the associated
+    //       Object3D's might be updated a few frames later after when it really need to be. It might
+    //       affect the user experience. Fix it if possible.
+    const observer = new IntersectionObserver(
+        (entries) => {
+            for (const entry of entries) {
+                // TODO: Endure the visibility set by other systems is not overridden. For example,
+                //       even if another system has made an entity visible within the panel for some reasons,
+                //       this system makes it visible. This issue should be fixed.
+                entry.target.object3D.visible = entry.intersectionRatio > 0;
+            }
+        },
+        {
+            root: root,
+            threshold: 0.0,
+        }
+    );
+    observer.observe(target);
+};
+
+/**
+ * @class BoundaryVisibilitySystem
+ * @classdesc Makes the entities invisible if they are outside of their parent panels
+ * @augments MRSystem
+ */
+export class BoundaryVisibilitySystem extends MRSystem {
+    /**
+     * @class
+     * @description BoundaryVisibilitySystem's default constructor.
+     */
+    constructor() {
+        super(false);
+        this.observedEntities = new WeakSet();
+    }
+
+    /**
+     * @function
+     * @description Called when a new entity is added to the scene.
+     * @param {MREntity} entity - the entity being added.
+     */
+    onNewEntity(entity) {
+        // TODO: Support nested panels
+        if (entity instanceof MRPanel) {
+            this.registry.add(entity);
+            entity.traverse((child) => {
+                if (child === entity) {
+                    return;
+                }
+
+                if (this.observedEntities.has(child)) {
+                    return;
+                }
+
+                this.observedEntities.add(child);
+                observe(entity, child);
+            });
+        } else if (!this.observedEntities.has(entity) && entity instanceof MRDivEntity) {
+            // There is a chance that a child entity is added after parent panel addition.
+            // Check registered panels and set up the observer if panels are found in parents.
+            for (const panel of this.registry) {
+                if (panel.contains(entity)) {
+                    this.observedEntities.add(entity);
+                    observe(panel, entity);
+                    break;
+                }
+            }
+        }
+    }
+}
--- a/src/core/componentSystems/BoundaryVisibilitySystem.js
+++ b/src/core/componentSystems/BoundaryVisibilitySystem.js
@@ -16,15 +16,24 @@ const observe = (root, target) => {
     const observer = new IntersectionObserver(
         (entries) => {
             for (const entry of entries) {
-                // TODO: Endure the visibility set by other systems is not overridden. For example,
-                //       even if another system has made an entity visible within the panel for some reasons,
-                //       this system makes it visible. This issue should be fixed.
+                // TODO: Ensure to avoid the visibility set up collision. If multiple systems set up the visibility
+                //       independently, only the last system before render call can have an effect. This callback
+                //       is fired asynchronously so it may be said that this callback is executed before all the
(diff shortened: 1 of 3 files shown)


src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,81 @@
...
+    //       visibility from the layout position changes. Therefore, the visibility of the associated
+    //       Object3D's might be updated a few frames later after when it really need to be. It might
+    //       affect the user experience. Fix it if possible.
+    const observer = new IntersectionObserver(
+        (entries) => {
+            for (const entry of entries) {
+                // TODO: Endure the visibility set by other systems is not overridden. For example,
+                //       even if another system has made an entity visible within the panel for some reasons,
+                //       this system makes it visible. This issue should be fixed.
+                entry.target.object3D.visible = entry.intersectionRatio > 0;
```
takahirox, line 22: Any good idea to fix this issue? @michaelthatsit @hanbollar

src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,81 @@
...
+        this.observedEntities = new WeakSet();
+    }
+
+    /**
+     * @function
+     * @description Called when a new entity is added to the scene.
+     * @param {MREntity} entity - the entity being added.
+     */
+    onNewEntity(entity) {
+        // TODO: Support nested panels
```
takahirox, line 54: I hope we do in future PRs

src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,81 @@
...
+ * @function
+ * @description Observe a target MRDivEntity and make the associated object visible only if it is in visible position in a root MRDivEntity
+ * @param {MRDivEntity} root
+ * @param {MRDivEntity} target
+ */
+const observe = (root, target) => {
+    // TODO: Callback is fired asynchronously so no guaranteed to be called immediately when the
+    //       visibility from the layout position changes. Therefore, the visibility of the associated
+    //       Object3D's might be updated a few frames later after when it really need to be. It might
+    //       affect the user experience. Fix it if possible.
```
takahirox, line 15: I'm happy if you folks test on headset and share thoughts @michaelthatsit @hanbollar

takahirox:
I'm happy if you folks help to test to check not to break other systems @michaelthatsit @hanbollar

src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,81 @@
...
+    );
+    observer.observe(target);
+};
+
+/**
+ * @class BoundaryVisibilitySystem
+ * @classdesc Makes the entities invisible if they are outside of their parent panels
+ * @augments MRSystem
+ */
+export class BoundaryVisibilitySystem extends MRSystem {
```
takahirox, line 38: Suggestion for better name ideas is welcome

michaelthatsit:
Tested in headset. the frame rate noticeably improves when models aren't visible. But when I scroll back down the models won't always reappear. see video and ignore my messy desk 😜

https://github.com/Volumetrics-io/mrjs/assets/19628643/04988cf2-8713-43ae-bfc4-0a173e9320f1

src/core/componentSystems/BoundaryVisibilitySystem.js
michaelthatsit: ROVSystem (Region of Visibility)?

src/core/componentSystems/BoundaryVisibilitySystem.js
michaelthatsit: oof. so this overrides the other systems? Calling it earlier in the Systems order might fix it. They're stepped sequentially, so any subsequent system would override this one.

takahirox:
@michaelthatsit Added a fallback. Can you test again to check it the problem has been resolved on your end, too?

src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,90 @@
...
+ * @description Observe a target MRDivEntity and make the associated object visible only if it is in visible position in a root MRDivEntity
+ * @param {MRDivEntity} root
+ * @param {MRDivEntity} target
+ */
+const observe = (root, target) => {
+    // TODO: Callback is fired asynchronously so no guaranteed to be called immediately when the
+    //       visibility from the layout position changes. Therefore, the visibility of the associated
+    //       Object3D's might be updated a few frames later after when it really need to be. It might
+    //       affect the user experience. Fix it if possible.
+    const observer = new IntersectionObserver(
```
takahirox, line 16: This technique (hack) using `IntersectionObserver` is almost same as the one used in #522. They may be combined.

takahirox:
If you think this optimization is not stable enough but the performance improvement is significant, we might think of this optimization as experimental for now and be able to introduce an opt-in flat maybe to `mr-app`.

src/core/componentSystems/BoundaryVisibilitySystem.js
hanbollar: PanelROVSystem? would prefer if we switch it to something that notes that it is panel specific - not hardlined about any of the other part of the naming

src/core/componentSystems/BoundaryVisibilitySystem.js
hanbollar: I think that delay is fine for now - as long as fixing the order per [link](https://github.com/Volumetrics-io/mrjs/pull/516/files#r1536268465) lets things run smoothly (ie the other systems can override / run what they need to) then it should be okay

src/core/componentSystems/BoundaryVisibilitySystem.js
```diff
@@ -0,0 +1,90 @@
...
+                //       independently, only the last system before render call can have an effect. This callback
+                //       is fired asynchronously so it may be said that this callback is executed before all the
+                //       sync systems.
+                entry.target.object3D.visible = entry.intersectionRatio > 0;
+            }
+            // Somehow callback is sometimes not fired even though crossing the threshold so as fallback
+            // always refresh the info as much as possible to keep it up-to-date.
+            // It seems that the callback is always called once soon after observe() is called,
+            // regardless of the intersection state of the entity.
+            // TODO: Confirm whether this behavior is intended. If it is not, there may be future
```
hanbollar, line 29: noting this todo as important

hanbollar:
added some comments as followup - left most of michael's comments as the main part of the review - pending that conversation i'll do a full review process

src/core/componentSystems/BoundaryVisibilitySystem.js
takahirox: As written in the updated inline comment, we may say that the callback is executed before all the sync systems because the callback is fired asynchronously.

But maybe it would be still a bit problematic in some scenarios. For example if a system updates a visibility of an object and want it to have an effect for some frames until certain condition is met, the callback of BoundaryVisibilitySystem can override the visibility.

PR #521, opened 2024-03-22
Fix MREntity.traverse()

takahirox:
## Linking

Fixes: https://github.com/Volumetrics-io/mrjs/pull/125/files#r1535082262

## Problem

MREntity.traverse() wants to go down to a child only if the child is MREntity. But due to a wrong check, it always goes down.

## Change for solution

Fix the bug. Currently `((!child) instanceof MREntity)` is used to decide not to go down but it would be always false because `!child` would be boolean (true/false). `!(child instanceof MREntity)` would be what it really needs to be.

Also, this commit includes a minor clean up.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/MREntity.js
+++ b/src/core/MREntity.js
@@ -413,11 +413,10 @@ export class MREntity extends MRElement {
         callBack(this);
         const children = Array.from(this.children);
         for (const child of children) {
-            // if o is an object, traverse it again
-            if ((!child) instanceof MREntity) {
-                continue;
+            // if child is an entity, traverse it again
+            if (child instanceof MREntity) {
+                child.traverse(callBack);
             }
-            child.traverse(callBack);
         }
     }
 


takahirox:
By the way, is it allowed that MREntities are placed under regular HTML elements like this?

```
<mr-div>
    <span>
        <mr-text>...</mr-text>
    </span>
</mr-div>
```

If yes, we want traverse() to go down to the leaves, and want it to call the callback only for MR entities, it would be move the traverse function from inside of MREntity to somewhere in utilities.

```javascript
const traverseEntities = (root, callback) => {
    const traverse = (entity) => {
        if (entity instanceof MREntity) {
            callback(entity);
        }
        for (let i = 0; i < entity.children.length; i++) {
            traverseEntities(entity.children[i], callback);
        }
    };
    traverse(root);
};
```

michaelthatsit:
> https://examples-mrjs-pr-521.onrender.com

Nah the expectation for now is that only mr-* tags will be within the mr-app tag. For simplicities sake.

michaelthatsit:
Tested in headset and this appears to work as expected.

michaelthatsit:
> By the way, is it allowed that MREntities are placed under regular HTML elements like this?
> 
> ```
> <mr-div>
>     <span>
>         <mr-text>...</mr-text>
>     </span>
> </mr-div>
> ```
> 
> If yes, we want traverse() to go down to the leaves, and want it to call the callback only for MR entities, it would be better to move the traverse function from inside of MREntity to somewhere in utilities.
> 
> ```js
> const traverseEntities = (root, callback) => {
>     const traverse = (entity) => {
>         if (entity instanceof MREntity) {
>             callback(entity);
>         }
>         for (let i = 0; i < entity.children.length; i++) {
>             traverseEntities(entity.children[i], callback);
>         }
>     };
>     traverse(root);
> };
> ```

Nah the expectation for now is that only mr-* tags will be within the mr-app tag. For simplicities sake.

michaelthatsit (approved):

PR #522, opened 2024-03-22
Optimize getBoundingClientRect()

### LINKED ISSUE #493: LayoutSystem seems to be slow

takahirox:
## Linking

Fixes #493 

## Problem

Element.getBoundingClientRect() is called from many places mainly to sync the layout between DOM elements and 3D scene. But .getCoundingClientRect() seems to be slow because it may invoke reflow.

## Change for solution

Add a hack with IntersectionObserver to avoid reflow in sync. Please refer to the inline code comments for the details.

## Notes

To reviewers: Please carefully read the inline comments about some concerns and decide whether to merge.

------------

## Required to Merge

- [ ] **BREAKING CHANGE**

### DIFF
--- a/src/core/MRElement.js
+++ b/src/core/MRElement.js
@@ -12,6 +12,45 @@ export class MRElement extends HTMLElement {
         super();
         this.environment = null;
         this.observer = null;
+
+        // Hack for the performance.
+        // Element.getBoundingClientRect() is called from many places
+        // mainly to sync the layout between DOM elements and 3D scene.
+        // But .getBoundingClientRect() is slow because it may invoke reflow.
+        // To avoid reflow in sync we use a hack with IntersectionObserver
+        // that gives .boundingClientRect in callback function.
+        //
+        // There are some concerns in this approach.
+        // 1. It might be costly to apply this technique to all MREntities.
+        //    Optimization might be necessary to apply it only to MREntities where
+        //    .getBoundingClientRect() is frequently executed.
+        // 2. Since boundingClientRect is updated asynchronously, the latest values
+        //    may be set a few frames late. It is necessary to confirm whether this
+        //    is acceptable for the user experience and various systems. One possible
+        //    workaround, which is not perfect but simple, would be to add a new method
+        //    that is more efficient but may not return the latest value, in addition to
+        //    the regular .getBoundingClientRect() method. The caller can then choose
+        //    which method to use depending on the purpose.
+        // 3. Asynchronous updates can make testing, problem investigation, and debugging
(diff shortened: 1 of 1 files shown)


michaelthatsit:
Hot damn! serious perf improvement in headset. Reading through the code now.

michaelthatsit:
I'm not too pumped about a hack, but I'm open to merging this one. Can you think of any better alternatives? would a modification to the LayoutSystem itself introduce a better solution?

michaelthatsit (approved):

takahirox:
If you think this optimization is not stable enough but the performance improvement is significant, we might think of this optimization as experimental for now and be able to introduce an opt-in flat maybe to mr-app.

takahirox:
> Can you think of any better alternatives?

`Element.getBoundingClientRect()` seems to be used a lot. Just to get width/height of an MR entity it seems to be called. If I'm right, somehow only one or few `Element.getBoundingClientRect()` calls in a frame seems to drop FPS counter significantly. So not only in `LayoutSystem` but from all other places we need to carefully manage not to call `Element.getBoundingClientRect()`.

An alternative in my mind is traversing up and calculating the layout position with parents `Element.offsetLeft` and others. I tried it but the calculated position didn't really match the result of `Element.getBoundingClientRect()`. Did I need to take into CSS account? Anyways, we may need more research if we want to go with this approach. If traversing up every time is costly, we may cache the result maybe per frame.

Also, it may be good to report to Meta for `Element.getBoundingClientRect()` performance issue in immersive mode. There might be a chance that the problem is Meta device platform specific.

hanbollar (approved):
code-wise marking as approved - deferring to the rest of @michaelthatsit @takahirox 's conversation for whether this pr should be merged or not

michaelthatsit:
I'm gonna merge this but created another issue to track the replacement of getBoundingRect

https://github.com/Volumetrics-io/mrjs/issues/528

PR #524, opened 2024-03-25
first pass fix for MRTextArea and MRTextField, allows for writing input

hanbollar:
## Linking

related to https://github.com/Volumetrics-io/mrjs/issues/513

## Problem

MRTextArea and MRTextField have been broken for a while (ie not finished) - completing part of their implementations in this pr - separating out arrow key movements and cursor movements fix into a different pr

## Solution

complete the implementations (`IMPORTANT NOTE:` - `this is pr 1 of a few to get this implemented`)

## Notes

adding a pr to documentation to comment out mrtextarea and mrtext field since they are and have been broken for awhile - will re-add them once this sequence of prs is fully pushed through and both are working again

`as of this pr pass:`
- mrtextarea input works
- mrtextfield input might work
- mrtextinput setup as base
- cursor needs work in both cases

------------

## Required to Merge

only need actions to pass since this is pr 1 of a few ~

- [x] **PASS** - all necessary actions must pass (excluding the auto-skipped ones)
~~- [ ] **TEST IN HEADSET** - [main dev-testing-example](https://github.com/Volumetrics-io/mrjs/tree/main/samples/index.html) and any of the other [examples](https://github.com/Volumetrics-io/mrjs/tree/main/samples/examples) still work as expected~~
~~- [ ] **VIDEO** - if this pr changes something visually - post a video here of it in headset-MR and/or on desktop (depending on what it affects) for the reviewer to reference.~~
- [x] **TITLE** - make sure the pr's title is updated appropriately as it will be used to name the commit on merge
- [x] **BREAKING CHANGE**
  - make a pr in the [documentation repo](https://github.com/Volumetrics-io/documentation) that updates the manual docs to match the breaking change
  - note: the docs underneath the `Javascript API` section come automated from the jsdocs inline with the code itself
  - link the pr of the documentation repo here: https://github.com/Volumetrics-io/documentation/pull/39
  - that documentation repo pr must be approved by `@lobau`

PR #527, opened 2024-03-26
update entity files: core-->core/entities && MR{name} --> MR{name}Entity

hanbollar:
## Links

Precursor to #524 

## Problem

too many 'entities that are and arent entities' in the pure core folder - making it odd for where to find things - pure core should be specifically overarching items `MREntity`, `MRElement`, `MRSystem`, and `MRApp` - everything else should be directly in its proper labeling folder

## Solution

moving the files and naming properly (removing dangling `Entity` from items)

### Breaking Change

This is not a breaking change from an html use-case perspective though it is a breaking javascript change

------------

## Required to Merge

~~- [ ] **VIDEO** - if this pr changes something visually - post a video here of it in headset-MR and/or on desktop (depending on what it affects) for the reviewer to reference.~~
~~- [ ] **BREAKING CHANGE**~~


michaelthatsit:
I think we should keep the `MR<name>Entity` system. IT follows the same style as HTMLElements (see screenshot and link, in the interfaces section under H)

https://developer.mozilla.org/en-US/docs/Web/API

Screenshot 2024-03-25 at 6 44 34 PM

hanbollar:
> I think we should keep the `MR<name>Entity` system. ...

I had removed the 'Entity' from the ends to rectify that we had so many entities that didnt end with 'Entity', ie just switched it to the more common version --> if you want them all to follow the pattern of `MR<name>Entity`, this also means MRMediaEntity, MRButtonEntity, MRTextEntity, etc

you good with that? if so, I'll update this pr to reflect that for all of them

michaelthatsit:
> > I think we should keep the `MR<name>Entity` system. ...
> 
> I had removed the 'Entity' from the ends to rectify that we had so many entities that didnt end with 'Entity', ie just switched it to the more common version --> if you want them all to follow the pattern of `MR<name>Entity`, this also means MRMediaEntity, MRButtonEntity, MRTextEntity, MRHyperLinkEntity, MRImageEntity, etc
> 
> you good with that? if so, I'll update this pr to reflect that for all of them

Yeah I'd rather do this. It's admittedly clunky, but it follows the design patterns we're attempting to inherit, given that these are still custom elements.

hanbollar:
> > > I think we should keep the `MR<name>Entity` system. ...
> > 
> > 
> > I had removed the 'Entity' from the ends to rectify that we had so many entities that didnt end with 'Entity', ie just switched it to the more common version --> if you want them all to follow the pattern of `MR<name>Entity`, this also means MRMediaEntity, MRButtonEntity, MRTextEntity, MRHyperLinkEntity, MRImageEntity, etc
> > you good with that? if so, I'll update this pr to reflect that for all of them
> 
> Yeah I'd rather do this. It's admittedly clunky, but it follows the design patterns we're attempting to inherit, given that these are still custom elements.

sg, updating now for that change ~

Issue #513, opened 2024-03-19
`mr-textarea` and `mr-textfield` are broken


lobau:
Simple case: https://codepen.io/lobau/pen/ExJZmKx
Or on the documentation https://docs.mrjs.io/doc/mr-textarea/

hanbollar:
updating for this to be a combined issue

PR #546, opened 2024-03-29
mrtextinput update pass 2 - separation into textarea and textfield

hanbollar:
## Linking

related to #524 and related to #513 

-----

all the rest of this section of the pr description is the same as in the #524 pr


hanbollar:
remaining for this task specifically: (other features can be added to it in future)
- [ ] finish making cursor x 'visually' match click selection loc (already directly matches when typing out)
- [ ] make cursor y 'visually' match click selection loc (already directly matches when typing out)
- [ ] handle the '\n' thing with enter making the slight offset at the end before new char
- [ ] cleanup: separate textfield out from textarea once the above is finalized

hanbollar:
> > remaining for this task specifically: (other features can be added to it in future)
> > 
> > * [x]   finish the left/right arrows
> > * [ ]   finish the up/down arrows
> > * [ ]   finish making cursor x 'visually' match click selection loc (already directly matches when typing out)
> > * [ ]   make cursor y 'visually' match click selection loc (already directly matches when typing out)
> > * [ ]   handle the '\n' thing with enter making the slight offset at the end before new char
> > * [ ]   cleanup: separate textfield out from textarea once the above is finalized
> 
> ## update on the above:
> * [x]  selections through clicking is implemented
> * [ ]  fix selection index is always zero
> * [x]  left arrows is implemented
> * [x]  right arrows is implemented
> * [ ]  fix right arrow has lingering 2step issue
> * [x]  up arrow is implemented
> * [x]  down arrow is implemented
> * [x]  up/down sometimes works sometimes doesnt
> * [ ]  still to fix: handle newLine wonkiness

## Remaining:

- [ ] fix selection index is always zero
- [ ] fix right arrow has lingering 2step issue
- [ ] handle new line wonkiness

hanbollar:
probably going to leave those unchecked in ## Remaining still as remaining off of this pr to get this in - and solve those in another pass

hanbollar:
for now cleaning up - to mrtextarea mrtextfield separation

hanbollar:
going to turn this into one more pr pass afterwards (the 3rd) that includes:
- the [remaining above needed 2 fixes](https://github.com/Volumetrics-io/mrjs/pull/546#issuecomment-2038832564) 
- some 'pretty-fication' for background to make them look more like actual 'input items'

@michaelthatsit adding you on review for a cursory look and awareness of this pr and its main setup logic
- in the next pr's review let's do a very in-depth look through since in that one things will be a lot more static
- and from that 3rd pr we can determine if it's ready to be exposed to users

hanbollar:
actually got in the small changes to make `check-format` happy - but ignore the rest of the documentation for those files, i'll actually make them nice in the next pr when things are finally solidifed

src/core/entities/MRTextAreaEntity.js
```diff
@@ -56,183 +41,124 @@ export class MRTextAreaEntity extends MRTextInputEntity {
...
     }
 
     /**
-     * Handles keydown events for scrolling and cursor navigation.
+     * Handles keydown events for scrolling and cursor navigation. Note
+     * that this is different than an input event which for our purposes,
+     * handles the non-navigation key-presses.
      * @param {event} event - the keydown event
      */
     handleKeydown(event) {
```
michaelthatsit, line 79: This really shouldn't be necessary if we're using the `hiddenInput`, right? you can get the selection (cursor position) directly from that.

hanbollar: hidden input doesnt handle the up/down keypresses as we would expect in a larger space - especially for textarea, it only treats them as left and right - hence the separation (see the comments in the `setupEventListeners` function)

that's something i'm going to simplify as part of the next pr given the lingering issue for '\n', but unfortunately for now both are still needed

michaelthatsit: I was using the keyboard events in the first version (prefunding), and it had quite a few issues, one being accessibility, as it doesn't launch the virtual keyboard.

input should work, I think you'd need to change it being `document.createElement('input')` to `document.createElement('textarea')` and that would give you the rows info.

michaelthatsit: Also events have a nasty habit of being caught out of order in some cases.

michaelthatsit (approved):
Approved based on our discussion. Let's find a way to manage cursor position without listening for `keydown` event by 0.7

PR #578, opened 2024-04-10
textInput pass 3 - adding some styling for background, fix cursor \n issue, and some more cleanup

hanbollar:
## Linking

related to #513 

pass1 was #524, pass2 was #546

-----

all the rest of this section of the pr description is the same as in the #524 / #546 prs

- [ ] fix selection index is always zero
- [x] consolidate input handling if possible (up/down cursor special case)
- [x] handle new line wonkiness
- [x] background color niceness for visual

------------

## Required to Merge

  - https://github.com/Volumetrics-io/mrjs/pull/578#issuecomment-2052704184
- [ ] **BREAKING CHANGE**


hanbollar:
pausing on this for a short bit

hanbollar:
todos:

- [ ] need good default options for the inputs
- [ ] selection index might be a dangling issue after this pr

hanbollar:
if all is good - then this is enough to re-expose this as a tag and in documentation repo with future issues created off of this for more work

hanbollar:
> if all is good - then this is enough to re-expose this as a tag and in documentation repo with future issues created off of this for more work

future niceties include:
- [ ] text change if a password input
- [ ] red outline as something people can toggle (troika has an easy outline feature)
- [ ] css change for if the field is disabled
- [ ] ...

michaelthatsit:
https://github.com/Volumetrics-io/mrjs/assets/19628643/2a550fcd-f3e0-435f-9fcf-09bf254d9f7b

michaelthatsit:
Still looking at the code but the cursor seems a little higher than it should be. Also did you test with other font sizes?

hanbollar:
> Still looking at the code but the cursor seems a little higher than it should be. Also did you test with other font sizes?

the cursor looks higher because im sort of faking the background height (the coloring one) to fit it decently with like a slight offset for niceties, but that nicety can probably be improved compared to the quick way im doing it - it matches the specific text location

and tested with other font sizes - it doesnt scale exactly as i want atm, but it does at least start at the right top and width position - was leaving that as a future noodle to not be a blocker on this one

src/core/entities/MRTextAreaEntity.js
```diff
@@ -12,56 +12,65 @@ import { MRTextInputEntity } from 'mrjs/core/entities/MRTextInputEntity';
...
         inputElement.style.position = 'absolute';
-        inputElement.style.left = '-9999px'; // Position off-screen
+        inputElement.style.left = '-9999px'; // Position off-screen so dont have weird white spot at top
         inputElement.style.height = '1px';
         inputElement.style.width = '1px';
         inputElement.style.overflow = 'hidden';
-        document.body.appendChild(inputElement); // Ensure it's part of the DOM for event capturing
+
+        // Ensure it's part of the DOM for event capturing
+        document.body.appendChild(inputElement);
```
michaelthatsit, line 39: I think we talked about this but why aren't we putting this in the shadow root? might scale a little better. especially for embedded experiences.

src/core/entities/MRTextFieldEntity.js
```diff
@@ -16,55 +14,91 @@ export class MRTextFieldEntity extends MRTextInputEntity {
      */
     constructor() {
         super();
-        this.wrapper = this.shadowRoot.appendChild(document.createElement('div'));
-        this.wrapper.innerHTML = '<slot></slot>';
+        this.lineHeight = 1.2; // Default line height, can be adjusted as needed
+        this.object3D.name = 'textField';
+
+        // this.wrapper = this.shadowRoot.appendChild(document.createElement('div'));
```
michaelthatsit, line 20: Related to my other comment. why a hidden field in the DOM instead of an actual field in the shadow?

src/core/entities/MRTextFieldEntity.js
```diff
@@ -73,7 +107,7 @@ export class MRTextFieldEntity extends MRTextInputEntity {
 
         // Ensure the cursor position is updated to reflect the current caret position
         setTimeout(() => {
```
michaelthatsit, line 109: what's this for?

src/core/entities/MRTextAreaEntity.js
hanbollar: oh no main reason - i can swap it to that

src/core/entities/MRTextFieldEntity.js
hanbollar: if we end up needing to pass the event info/etc to the function within that, we want the timeout to have a delay s.t. the event actually holds all the proper information (since it's a press/trigger type one) otherwise, the events can get delayed so the next time you grab the info, you're grabbing the previous version until the next event gets called (it's like the stack gets delayed by one or more steps if that makes sense)

---

tldr atm this setup is no longer needed since i dont pass the 'event' object to the cursor function anymore, but if we ever do it again, it will be needed with something > 0

^ given this, lmk if you'd want it removed or kept just in case

src/core/entities/MRTextAreaEntity.js
michaelthatsit: Yeah it shouldn't be a huge issue and then you wouldn't need to do the 9999999/-99999999 absolute positions. you can just set it to `display:none` in CSS.

hanbollar: that is so much cleaner haha

src/core/entities/MRTextFieldEntity.js
michaelthatsit: I'd say if we're not using it we can lose it for the sake of cleanliness

src/core/entities/MRTextAreaEntity.js
michaelthatsit: Yeah use this as an example https://github.com/Volumetrics-io/mrjs/pull/578#discussion_r1563401938

It's one of those stack overflow tricks I picked up the first time I was trying with MRTextField.

michaelthatsit:
Overall it's pretty solid. I don't think you're blocking anything ATM so try and make the Shadow DOM changes we discussed. 

Some notes for what I think we need by v1:

- [ ] select cursor position (blocked by click events I think)
- [ ] Text selection
- [ ] copy/paste
- [ ] bonus: spell check

michaelthatsit:
Screenshot 2024-04-12 at 5 31 19 PM

Accidental caps but just noticed that text doesn't stay in the textarea if you type too far.

hanbollar:
> Screenshot 2024-04-12 at 5 31 19 PM
> Accidental caps but just noticed that text doesn't stay in the textarea if you type too far.

yupyup - one of the few task for this was to add in a scrolling offset - so that the number of lines are only limited to what can show up in the main area but the text can be handled properly, so i didnt make a change to limit the height of it based on text (we also cant do mask of a mask either, so that wouldnt hv been worth it)

if that's a blocker for this pr lmk, but i had added in the todos within the code as a task after this pr

michaelthatsit:
No blocker but let's prioritize it on the next pass. Good work!

PR #599, opened 2024-04-30
add `loop` and `loopMode` capabilities for AnimationSystem, fix for `comp-audio` to match `comp-anim`: `state`-->`action`

hanbollar:
## Linking

related to:
- [loop not impl](https://discord.com/channels/1137890872688660500/1219366027545149462/1234669764811558943)
- [audio `state` --> `action` for consistency](https://discord.com/channels/1137890872688660500/1219366027545149462/1234671001841700934)

## Problem

loop not implemented for data-comp-animation

## Solution

implement it

### Breaking Change

YES!!
- changes audio state --> `action`. Made sure this was updated in the documentation repo and the samples here
- adds looping for data-comp-animation

## Notes

...

------------

## Required to Merge

https://github.com/Volumetrics-io/mrjs/assets/19377312/799ab370-5f30-412a-8643-59b4241214c5

- [x] **BREAKING CHANGE**
    - link the pr of the documentation repo here: https://github.com/Volumetrics-io/documentation/pull/50


src/core/componentSystems/AnimationSystem.js
```diff
@@ -101,24 +103,95 @@ export class AnimationSystem extends MRSystem {
...
-                entity.mixer.clipAction(clip).pause();
-                break;
-            case 'stop':
-                entity.mixer.clipAction(clip).stop();
-                break;
-            default:
-                break;
+        let action = entity.mixer.clipAction(clip);
+
+        if (comp.hasOwnProperty('action')) {
```
michaelthatsit, line 115: You can also do `comp?.action` which is a little cleaner.

michaelthatsit (commented):
Looks great with a few comments!

src/core/componentSystems/AnimationSystem.js
```diff
@@ -101,24 +103,95 @@ export class AnimationSystem extends MRSystem {
...
+                    action.stop();
+                    break;
+                default:
+                    mrjsUtils.error.err('Unknown case hit for action in the AnimationSystem from entity:', entity, '. Comp is:', comp);
+                    return;
+            }
+        }
+
+        let hasLoop = comp.hasOwnProperty('loop');
+        let hasLoopMode = comp.hasOwnProperty('loopMode');
```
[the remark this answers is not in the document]
hanbollar: that property allows for 'once', 'repeat', and 'pingpong'

was trying to mirror it with `loop` to help with the pairing/that they go together (`loop` && `loopMode`)

i can switch it to `repeatMode`? but i'm not sure that does well from a phrasing perspective with `loop` and the fact that it allows for `repeat` and `pingpong`

lmk your thoughts ~

src/core/componentSystems/AnimationSystem.js
hanbollar: i like the simpler approach you note, but that also makes it turn into a truthy check of the `action` value if it exists, meaning it can silently error (instead of hitting the default case) if a user makes a typo and leaves it as a 0 or an empty string for example

i'd prefer to keep it as the hasOwnProperty line, but lmk your thoughts

src/core/componentSystems/AnimationSystem.js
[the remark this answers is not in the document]
michaelthatsit: ah ok. It's more out of an afinity for single word properties. but loopMode makes more sense.

src/core/componentSystems/AnimationSystem.js
michaelthatsit: hmm does comp?.action == null not hit the default state?

michaelthatsit:
Approved for merge. 

I use `object.property` pretty much everywhere so we miiight need to decide which approach to take across the board for consistency at some point.

src/core/componentSystems/AnimationSystem.js
hanbollar: sorry im confused - all that does is check if comp exists - and then whether comp.action == null

in the `comp.hasOwnProperty` line im trying to check if action exists before running through our switch statements - do you have a different preferred way of writing that?

in the audio system it looks like that's setup as `comp.action ?? false` - is that what you'd prefer instead?

hanbollar: i think i'd still be a little worried of the `comp.action` value existing but being incorrect as a truthy-ness option and not going into if-statement 🤔 

thoughts?

hanbollar:
> I use `object.property` pretty much everywhere so we miiight need to decide which approach to take across the board for consistency at some point.

oh i think `object.property` is fine most places we have it - i just think this is an edge situation where we need to be very careful of user entry given the multiple options that arent just true/false

src/core/componentSystems/AnimationSystem.js
hanbollar: i'm going to merge it in since approved so not blocker, but lmk since that's a quick pr fix afterwards if needed

PR #619, opened 2024-05-07
noodle - inheritance not being handled right - needs documentation update for expln

### LINKED ISSUE #618: children of a custom entities are not visible
If you add children to a custom entity you define as extending MREntity, it's DOM children are not visible in the MRjs scene.

For example, in this sample code:
https://codepen.io/lobau/pen/wvZVLw

```html
<mr-app orbital="true">
    <!-- Left one -->
    <mr-div data-position='-0.3 0 0'>
        <mr-model src='https://assets.codepen.io/686746/koi_fish.glb' style="scale: 0.1"></model>
    </mr-div>
    
    <!-- Right one -->
    <mr-foo data-position='0.3 0 0'>
        <mr-model src='https://assets.codepen.io/686746/koi_fish.glb' style="scale: 0.1"></model>
    </mr-foo>
    
</mr-app>
```

The defined custom element, `mr-foo`:

```js
class MRFoo extends MREntity {

    constructor() {
        super()
    }

    connected() {
        this.boxMesh = new THREE.Mesh(
            new THREE.BoxGeometry(0.2, 0.2, 0.2),
            new THREE.MeshPhongMaterial({
                color: "#ff9900",
                transparent: true,
                opacity: 0.4,
            }));
        this.object3D.add(this.boxMesh);
    }
    
}

customElements.define('mr-foo', MRFoo);
```

On the left, you have an `mr-model` as a child of an `mr-div` and it works as expected.
On the right, you have a custom element (`mr-foo`) that is defined as `class MRFoo extends MREntity`. The child `mr-model` is not on the MRjs scene.

image

My assumption is that it has to do with adding the slotted elements in the shadowDOM maybe?

hanbollar:
## Linking

Fixes #618 

## Problem

Screenshot 2024-05-07 at 3 29 00 PM

(see the rest in the issue description)

something is wrong with inheritance for user designated items

## Solution

*Quick explanation of change to be done*

needs more noodling first

### Breaking Change

*If this is a breaking change describe the before and after and why the change was necessary*

## Notes

*Notes and any associated research or links*

------------

## Required to Merge

- [ ] **BREAKING CHANGE**


hanbollar:
### The issue

It has to do with mr.js taking longer to load - so even though mr.js /isnt/ an async load and you /should/ be able to do :

```
<script src="./mr.js">
<script>
  class MRFoo extends MREntity {
     ...
  }
</script>
```

which gives us the sad state where the model doesnt load under `MRFoo`

Screenshot 2024-05-10 at 7 15 38 PM

which in our case means, `MREntity` finally becomes defined after `MRFoo` is already being handled, which is /not/ what we want.

Screenshot 2024-05-10 at 7 15 49 PM

-------

### The reasoning and short term fix

mrjs is big enough that it takes some time for its items to load one by one - which means unfortunately we end up with a need to have people add items to extend through `async` if they want mrjs to be handled 100% as expected

that is, at the moment we need them to do something like:

```
<!-- MRjs -->
<script>
function defineMRFoo() {
  class MRFoo extends MREntity {
     ...        
  }
</script>
<script src="./mr.js" onload="defineMRFoo()"></script>
```

which means we get what we wanted:

Screenshot 2024-05-10 at 7 03 36 PM

-----

### MRFoo

note: we currently define `MRFoo` as follows:

```
class MRFoo extends MREntity {
    constructor() {
        super();
    }

    connected() {
        super.connected();
        this.boxMesh = new THREE.Mesh(
            new THREE.BoxGeometry(0.2, 0.2, 0.2),
            new THREE.MeshPhongMaterial({
                color: "#ff9900",
                transparent: true,
                opacity: 0.4,
            }));
        this.object3D.add(this.boxMesh);
    }
}

// Define custom element
customElements.get('mr-foo') || customElements.define('mr-foo', MRFoo);
```

hanbollar:
@lobau @michaelthatsit let me know if you have any experience with delayed loading of libraries for handling in the html directly - i dont really want to force users to do an `onload` function to extend anything from our `mr.js` lib

hanbollar:
talk with @michaelthatsit 

some things:
- try `async connected` and adding `await` in front of `super.connected()` ??
- look into the add function as well, maybe queue up and wait for load before checking validity
- maybe remove add and have that be function that extends `appendChild`

hanbollar:
> * try `async connected` and adding `await` in front of `super.connected()` ??

doesnt work - hitting same issue of :
Screenshot 2024-05-13 at 11 20 40 AM

hanbollar:
double `defer` doesnt work - takes too long too load

Screenshot 2024-05-13 at 11 49 34 AM

hanbollar:
Screenshot 2024-05-13 at 12 53 49 PM

It seems to work ok if we have src for mrjs script in header and the class definition in the header

this is because we are using mrfoo as an html element within the <body> group and then defining it after - so all the asyncs we're doing for the scripts wont work because of the html tag predefinition. The `onload` function version works from https://github.com/Volumetrics-io/mrjs/pull/619#issuecomment-2105457895 since it blocks everything related until after.

What i find is interesting is that adding `defer` to the source script in header of `mrjs` and the creation script of `class MRFoo` which is after the body, doesnt have the same effect as the `onload` function does. But from what I understand, that's because the `onload` function links the two together so `MRFoo` still ends up being created fully before the `body`'s html information gets executed.

hanbollar:
## Solved!

### Starting Notes

@lobau @michaelthatsit - it seems to work all fine if we have users define things in the header after the <script src=mrjs> tag - it seems the reason for the issue was:
- even though we can do async etc of `mrjs` loading in header with `mrfoo` in a script after `<mr-app>` is created, the `onload` function is needed as part of <script src=`mrjs`> to connect `mr-foo` to that part of the loading before any of the <body> tag elements are included

### The Cause

that is, this is what was causing the issue

```
<header>
    <script src="./mr.js"></script>
</header>
<body>
    <mr-app>
         ...
         <mr-foo>...</mr-foo>
         ...
    </mr-app>
    <script>
          ...
          class MRFoo extends MREntity { ... }
          ...
    </script>
</body>
```

so the `<mr-foo>` tag would be hit since it's not on async or defer, even if the import script for `mrjs` and the construction script for `MRFoo` have `defer` in their tag.

doing the `onload` for `<script src="./mr.js" onload="constructMRFoo()"></script>` (that was noted in the [previous comment](https://github.com/Volumetrics-io/mrjs/pull/619#issuecomment-2105457895)) works just because it ties `MRFoo`'s implementation timing to `mrjs`'s timing before the rest of the body gets fully filled out (even though it's already read)

### The Fix

to not require that, we just need to do the following to make sure the `<mr-foo>` html element inside `<mr-app>` is happy just like everything else:

```
<header>
    <script src="./mr.js"></script>
    <script>
          ...
          class MRFoo extends MREntity { ... }
          ...
    </script>
</header>
<body>
    <mr-app>
         ...
         <mr-foo>...</mr-foo>
         ...
    </mr-app>
</body>
```

hanbollar:
going to close this as solved but not merging since this was to figure out why the problem is happening and what to recommend to users going foward - now the noodle is complete 🥳

